PR #1, opened 2016-08-30
Add a main module file and a module logo

nimid:
**1. Objective reason**

To install the module in PrestaShop, it requires a main file that contains a main class and meets [the module installation requirements](http://doc.prestashop.com/display/PS16/Creating+a+first+module#Creatingafirstmodule-Creatingafirstmodule).

The declarations in file, omise.php, enable the module to install in PrestaShop.

Related ticket: T781

**2. Description of change**

The user can install the module in their PrestaShop web site. The module still has no any functions. The user will see the Omise logo at the modules list.

The screenshot below shows that the main module file can enables the module installation in PrestaShop.

screenshot-localhost 2016-08-30 13-56-36

All

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

`-`


omise/omise.php
```diff
@@ -0,0 +1,44 @@
+<?php
+if (! defined('_PS_VERSION_'))
+    exit();
+
+class Omise extends PaymentModule
+{
+
```
guzzilar: @nimid can you remove this line break? thanks

omise/omise.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * The version of the module.
+     *
+     * @var string
+     */
+    const MODULE_VERSION = '1.6.0.0';
+
+    public function __construct()
+    {
+        $this->name = self::MODULE_NAME;
```
guzzilar: @nimid from the code style, I'd prefer to make equal sign align in the same column as much as possible,
Do you mind if I'll update them to something like below.

``` php
public function __construct()
{
    $this->name                   = self::MODULE_NAME;
    $this->tab                    = 'payments_gateways';
    $this->version                = self::MODULE_VERSION;
    $this->author                 = 'Omise';
    $this->need_instance          = 0;
    $this->ps_versions_compliancy = array('min' => '1.6', 'max' => '1.6');
    $this->bootstrap              = true;

    parent::__construct();

    $this->displayName            = self::MODULE_DISPLAY_NAME;
    $this->confirmUninstall       = $this->l('Are you sure you want to uninstall ' . self::MODULE_DISPLAY_NAME . ' module?');
}
```

guzzilar:
I'm ok for the rest. 👍

guzzilar:
tested on 1.6.1.7
screen shot 2559-09-14 at 6 03 15 pm

👍

omise/omise.php
nimid: @guzzilar I disagree. There are no PHP coding standard mention or encourage this style. I found no declaration for this style to be should or must. Including, PrestaShop 1.6 and 1.7 did not use this style also.

To be consistent this style, it give more work to other developer.

For an example, if one of them would be named longer or shorter, all involved variables need to be aligned variables with it.

This style requires human do it manually. I found no any automated tool to do this style.

Although it is the benefit of this style but it is unnecessary in this point.

guzzilar: @nimid I don't think so. those standard didn't say doesn't mean we should not do it (unless it gonna break something or some rules that those standard had set).

To support code readability, I think alignment of assignments is also important.
But anyway, we can say 👆 style could be broken when the length of the variable makes another surrounding variable looks strange.

As for the automated tool, you can search for "alignment of assignments" keyword,
I did and found tools like that for `PHPStrom`, `Sublime3` & `NetBeans`

for example, https://www.youtube.com/watch?v=_c6ft-miHr4

actually, it's not about rule, just code style, and we can set whatever style we want together (just at least, don't break the rule from PSR-2). :D

don't you think above code is more readable?

guzzilar: btw, I'll do that for only variables are in the same group and only adjust just for the inside method only

i.e.

``` php
public function __construct()
{
    $this->name                   = self::MODULE_NAME;
    $this->tab                    = 'payments_gateways';
    $this->version                = self::MODULE_VERSION;
    $this->author                 = 'Omise';
    $this->need_instance          = 0;
    $this->ps_versions_compliancy = array('min' => '1.6', 'max' => '1.6');
    $this->bootstrap              = true;

    parent::__construct();

    $this->displayName            = self::MODULE_DISPLAY_NAME;
    $this->confirmUninstall       = $this->l('Are you sure you want to uninstall ' . self::MODULE_DISPLAY_NAME . ' module?');
}

// I will not try to make alignment of this method to equal above
public function something()
{
    $this->name  = 'Salmon';
    $this->type  = 'Fish'
    $this->price = '99 USD';
}
```

``` php
public function something()
{
    $this->name  = 'Salmon';
    $this->type  = 'Fish'
    $this->price = '99 USD';

    // I will not try make alignment for
    // these 2 blocks of code, cuz it's too much and they are difference group.
    if (something) {
        $new_name  = "Unagi";
        $new_type  = "maybe fish?";
        $new_price = "200 USD";
    }
}
```

nimid: Readability is subjective. For me both styles have readability.

That video is nice but I did not use Sublime3 and how about other developers use other text editor or IDE.

I care about other developers that need to consistent this extended style.

This point is very small but if it matter to you, you can change it.

nimid: @guzzilar If this style was introduced, it required the maintenance and consistency all files in this project. Although, this time there are not many files but in the future, this project will growing definitely.

Form my opinion, we are slow in the process of code review especially the point and discussion like this.

In the code review, I like to found the problem, non-function, or bug so that it can be removed before release to the user.

I encourage to use a tool to automatically check the standard of the source code such as PHP_CodeSniffer or other tools that can be run on the developer machine (at the first step) and it can reduce the time, objective and discussion in part of code review.

I strongly agree with coding standard but if it slow the time to market down, it must be carefully considered.

guzzilar: @nimid 
Readability is subjective. For me both styles have readability.
☝️ yeah, back to, if you don't mind I'll change it for you.

for IDE, it's just `space character`, I don't see any issue that make another IDE or text editor can't follow that style.

for review process, understand that.

guzzilar: to be quick, I'll update it and let you review & merge it now

guzzilar:
@nimid pls review https://github.com/omise/omise-prestashop/pull/1/commits/ae13ede15c34c6f78447b090f4163126821fb84a & merge it if you are ok with.

PR #3, opened 2016-08-30
Initial the unit testing and add a unit test for the main class, Omise

nimid:
**1. Objective reason**

To deliver quality software, the unit testing is one of many parts that can be achieved. It is easy to initial the unit testing at the beginning of the development.

Required pull request: [#1](https://github.com/omise/omise-prestashop/pull/1)

**2. Description of change**

Initial the unit testing and add a unit test for the main class, Omise. The unit testing will be run with PHPUnit.

Note:
1. To run unit test
   
   `$ phpunit`
2. To run unit test with code coverage reporting
   
   `$ phpunit --coverage-html report/`

From the second sample command, the report will be generated to the directory `report/`. It can be changed to another directory.

None

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

`-`

### DIFF
--- a/tests/unit/OmiseTest.php
+++ b/tests/unit/OmiseTest.php
@@ -1,0 +1,37 @@
+<?php
+define('_PS_VERSION_', 'TEST_VERSION');
+
+class OmiseTest extends PHPUnit_Framework_TestCase
+{
+    private $omise;
+
+    public function setup()
+    {
+        $paymentModule = $this->getMockBuilder(stdClass::class)
+            ->setMockClassName('PaymentModule')
+            ->setMethods(array('__construct', 'l'))
+            ->getMock();
+
+        $this->omise = new Omise();
+    }
+
+    public function testName_omise()
+    {
+        $this->assertEquals('omise', $this->omise->name);
+    }
+
+    public function testDisplayName_Omise()
+    {
+        $this->assertEquals('Omise', $this->omise->displayName);
+    }
+
+    public function testNeedInstance_0()
+    {
+        $this->assertEquals(0, $this->omise->need_instance);
+    }
+
+    public function testBootstrap_true()
+    {
+        $this->assertEquals(true, $this->omise->bootstrap);
+    }
+}
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -8,4 +8,9 @@
             <directory>tests/unit/</directory>
         </testsuite>
     </testsuites>
+    <filter>
+        <whitelist>
+            <directory suffix=".php">omise</directory>
+        </whitelist>
+    </filter>
 </phpunit>
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/5.5/phpunit.xsd"
+    bootstrap="tests/autoload.php">
+
+    <testsuites>
+        <testsuite name="Unit">
+            <directory>tests/unit/</directory>
+        </testsuite>
+    </testsuites>
+</phpunit>
--- a/tests/autoload.php
+++ b/tests/autoload.php
@@ -1,0 +1,14 @@
+<?php
+function autoload($class) {
+    static $classes = null;
+
+    if ($classes === null) {
+        $classes = array('Omise' => 'omise.php');
+    }
+
+    if (isset($classes[$class])) {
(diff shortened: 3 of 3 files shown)


tests/unit/OmiseTest.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        $paymentModule = $this->getMockBuilder(stdClass::class)
+            ->setMockClassName('PaymentModule')
+            ->setMethods(array('__construct', 'l'))
+            ->getMock();
+
+        $this->omise = new Omise();
+    }
+
+    public function testName_omise()
```
guzzilar, line 18: @nimid for the naming of tests, it's better to explicit your test behaviour by have `test scenario` & `expectation` in the test name, and use english sentence pattern there, like
`testThatTheOmiseNamePropertyShouldReturnOmise`

Btw, some pattern says, you should have `test scenario` at the name, some pattern says you should have both `test scenario` & `expectation` at the name (and some says, use english sentence like ,"should", "must", etc in the test) to make it clear and it give us a benefit that we can easily understand the test by reading from the test name when it fails.

For me, I'm fine both, 
just it should be understandable when we skim through the lines without check the code to understand what's the test aim to prove.

Btw, I usually do like

``` php
/**
 * @test
 */
public function people_should_happy_when_they_get_dog
{
    // assertion.
}

// or

/**
 * @test
 */
public function someone_must_feed_the_dog_food_when_they_bark
{
    // assertion.
}
```

I usually move `test` prefix to `@ test` annotation and use `_` to concat the string to make it easy to read and looks like english sentence instead of name it with `testCamelCase`.

nimid (commented):
@guzzilar In the unit testing, the name of test case I use the style of  3 compartments. It consists of:
1. The name of method or attribute under test
2. The condition or scenario
3. The expectation

They were named by camel case and separated by an underscore.

For an example from the [test cases that I have committed](https://github.com/omise/omise-prestashop/blob/9f8f29d2310da44322f49e111db172395607ec28/tests/unit/OmiseTest.php), 

``` php
public function testHookPayment_moduleIsActivatedAndEnabled_displayThePaymentForm()
        {
            $this->omise->active = true;
            $this->setting->method('isModuleEnabled')->willReturn(true);
            $this->omise->method('display')->willReturn('payment_form');

            $this->assertEquals('payment_form', $this->omise->hookPayment(''));
        }
```

I agree with the point that the test case name should has more descriptive and explanation for other developers.

Changing the test case name will be in the next pull request.

nimid:
Test

tests/unit/OmiseTest.php
nimid: Test

guzzilar:
@nimid got it,
anyway, could you consider about both name below,

``` sh
~/D/O/omise-prestashop (feature/initial-unit-test) $ ./vendor/bin/phpunit tests
PHPUnit 4.4.5 by Sebastian Bergmann.

Configuration read from /omise-prestashop/phpunit.xml

...F.F

Time: 51 ms, Memory: 3.50MB

There were 2 failures:

1) OmiseTest::HookPayment_module_is_activated_and_enabled_display_title
Failed asserting that true matches expected 'x'.

/omise-prestashop/tests/unit/OmiseTest.php:38

2) OmiseTest::HookPayment_module_is_activated_and_enabled_display_the_payment_form
Failed asserting that true matches expected 'x'.

/omise-prestashop/tests/unit/OmiseTest.php:51

FAILURES!
Tests: 6, Assertions: 6, Failures: 2.
~/D/O/omise-prestashop (feature/initial-unit-test) $ 
```

and

``` sh
~/D/O/omise-prestashop (feature/initial-unit-test) $ ./vendor/bin/phpunit tests
PHPUnit 4.4.5 by Sebastian Bergmann.

Configuration read from /omise-prestashop/phpunit.xml

...F.F

Time: 50 ms, Memory: 3.50MB

There were 2 failures:

1) OmiseTest::testHookPayment_moduleIsActivatedAndEnabled_displayTitle
Failed asserting that true matches expected 'x'.

/omise-prestashop/tests/unit/OmiseTest.php:36

2) OmiseTest::testHookPayment_moduleIsActivatedAndEnabled_displayThePaymentForm
Failed asserting that true matches expected 'x'.

/omise-prestashop/tests/unit/OmiseTest.php:46

FAILURES!
Tests: 6, Assertions: 6, Failures: 2.
```

I like the first one because it's kind of I'm reading just sentence. What do you think?

Btw, I'm ok with your naming style, feel free to ignore above. :)

nimid:
@guzzilar For the test case name that consist of scenario and expectation, using the underscore to be the separator. It is not easy to distinguish.

PR #5, opened 2016-09-06
Display the initial setting page

nimid:
**1. Objective reason**

Display the controls at the setting page.

It is the display only. It has not been any function.

Required pull request: [#1](https://github.com/omise/omise-prestashop/pull/1)

**2. Description of change**

Implement a function, Omise.getContent(). Define the parameters for display the controls.

The screenshot below shows the setting page.

screenshot-localhost 2016-09-10 16-46-35

All

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

`-`

### DIFF
--- a/omise/omise.php
+++ b/omise/omise.php
@@ -47,82 +47,82 @@ class Omise extends PaymentModule
             'legend' => array(
                 'title' => $this->l('Settings')
             ),
-            'input' => array(
+            'input'  => array(
                 array(
-                    'type' => 'switch',
-                    'label' => $this->l('Enable/Disable'),
-                    'name' => 'module',
-                    'is_bool' => true,
-                    'desc' => $this->l('Enable Omise Payment Module.'),
-                    'values' => array(
+                    'type'     => 'switch',
+                    'label'    => $this->l('Enable/Disable'),
+                    'name'     => 'module',
+                    'is_bool'  => true,
+                    'desc'     => $this->l('Enable Omise Payment Module.'),
+                    'values'   => array(
                         array(
-                            'id' => 'module_enabled',
+                            'id'    => 'module_enabled',
                             'value' => 1,
                             'label' => 'Enabled'
                         ),
                         array(
-                            'id' => 'module_disabled',
+                            'id'    => 'module_disabled',
                             'value' => 0,
                             'label' => 'Disabled'
                         )
                     )
                 ),
                 array(
-                    'type' => 'switch',
-                    'label' => $this->l('Sandbox'),
-                    'name' => 'sandbox',
-                    'is_bool' => true,
-                    'desc' => $this->l('Enabling sandbox means that all your transactions will be in TEST mode.'),
-                    'values' => array(
+                    'type'     => 'switch',
+                    'label'    => $this->l('Sandbox'),
+                    'name'     => 'sandbox',
+                    'is_bool'  => true,
+                    'desc'     => $this->l('Enabling sandbox means that all your transactions will be in TEST mode.'),
+                    'values'   => array(
                         array(
-                            'id' => 'sandbox_on',
+                            'id'    => 'sandbox_on',
                             'value' => 1,
                             'label' => 'Enabled'
                         ),
                         array(
-                            'id' => 'sandbox_off',
+                            'id'    => 'sandbox_off',
                             'value' => 0,
                             'label' => 'Disabled'
                         )
                     )
                 ),
                 array(
-                    'type' => 'text',
-                    'label' => $this->l('Public key for test'),
-                    'name' => 'publicKeyForTest',
+                    'type'     => 'text',
+                    'label'    => $this->l('Public key for test'),
+                    'name'     => 'publicKeyForTest',
                     'required' => false,
-                    'desc' => 'The "Test" mode public key can be found in Omise Dashboard.'
+                    'desc'     => 'The "Test" mode public key can be found in Omise Dashboard.'
                 ),
                 array(
-                    'type' => 'text',
-                    'label' => $this->l('Secret key for test'),
-                    'name' => 'secretKeyForTest',
+                    'type'     => 'text',
+                    'label'    => $this->l('Secret key for test'),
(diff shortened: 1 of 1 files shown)


guzzilar:
rebased

guzzilar:
@nimid I updated code style here, https://github.com/omise/omise-prestashop/pull/5/commits/a938f18823f0f1ace0e62a1ce899e47c1e452692 pls review

omise/omise.php
```diff
@@ -40,4 +40,100 @@ public function __construct()
...
+                ),
+                array(
+                    'type'     => 'text',
+                    'label'    => $this->l('Public key for test'),
+                    'name'     => 'publicKeyForTest',
+                    'required' => false,
+                    'desc'     => 'The "Test" mode public key can be found in Omise Dashboard.'
+                ),
+                array(
+                    'type'     => 'text',
```
guzzilar, line 97: @nimid pls change this field to `password` type

omise/omise.php
```diff
@@ -40,4 +40,100 @@ public function __construct()
...
+                    'label'    => $this->l('Public key for live'),
+                    'name'     => 'publicKeyForLive',
+                    'required' => false,
+                    'desc'     => 'The "Live" mode public key can be found in Omise Dashboard.'
+                ),
+                array(
+                    'type'     => 'text',
```
guzzilar, line 111: @nimid pls change this field to password type

guzzilar (changes requested):

nimid (commented):
@guzzilar It is ok.

nimid:
@guzzilar `a938f18` is ok.

omise/omise.php
nimid: @guzzilar If this type was changed from text to password, the default PrestaShop password type is different from text.

It style was changed and its value can not be show from the setting.

screenshot-localhost 2016-09-16 11-13-54

What is the reason for this type to be password?

guzzilar: @nimid ah, the original is, follows from Omise-WooCommerce https://github.com/omise/omise-woocommerce/blob/master/includes/omise-wc-setting.php#L36
and the reason that I changed at Omise-WooCommerce's field type to password because `secret_key` is confidential information. and WordPress can have multiple admin access the admin page. So, I don't want all of admin can see secret key.

nimid: @guzzilar Ok.

This setting page was automatically generated by the PrestaShop function by using the array configuration.

If the secret key need to be password type, using the generator function of PrestaShop is not appropriate.

It need to use the template file. That is all the implementation to display the setting page will be changed.

So, to change the secret key to be password type, it should be in the next pull request.

guzzilar: @nimid  ✋ I'm ok to wait for another PR.
But just curious, why we have to shift it to another PR? 
Cuz the PR topic is about implement setting page, and if you gonna change `secret key` you have to reimplement setting page again, so why we can't finish in this PR?

nimid: @guzzilar For PrestaShop, I found it already have the feature to manage the permission for each group of user (profile).

screenshot-localhost 2016-09-16 11-40-24

What do you think?

guzzilar: Nice to have (and I like it), but afraid that it's gonna take time for the first release.

In Omise Dashboard, we will remove `secret key (live)` from the keys page and users will not be able to see it anymore after 15 mins that they roll new key. So, I think we can also hide it from the plugin setting page too.

Btw, what is your concern to don't wanna use `array configuration` to config field type to `password`, is it about UI design?

nimid: @guzzilar The UI does not matter but the behavior is the concern.

After save the secret key that is the password type, the normal behavior of PrestaShop it will not retrieve from the setting.

It means if the user open the page again the value in the secret key is empty.

So to change the secret key from the text type to pass word type, it need to implement the template file not the array configuration.

I have researched when I do this feature but can not found. If you can find, how to easily show the value in password type with the array configuration. It is very nice.

guzzilar: @nimid got it,
also just checked from Prestashop code,

``` html
<div class="input-group fixed-width-lg">
    <span class="input-group-addon">
        <i class="icon-key"></i>
    </span>
    <input type="password"
        id="{if isset($input.id)}{$input.id}{else}{$input.name}{/if}"
        name="{$input.name}"
        class="{if isset($input.class)}{$input.class}{/if}"
        value=""
        {if isset($input.autocomplete) && !$input.autocomplete}autocomplete="off"{/if}
        {if isset($input.required) && $input.required } required="required" {/if} />
</div>
```

it's impossible to retrieve data and assign to the `password` field's value with that array configuration, they left `value` attr as `value=""` 😶

can I jump back to my previous question, 
"But just curious, why we have to shift it to another PR? 
Cuz the PR topic is about implement setting page, and if you gonna change secret key you have to reimplement setting page again, so why we can't finish in this PR?"

(Btw, for me, implement with their role/permission system is good idea, but just afraid that it gonna take time for first release.)

nimid: @guzzilar What do you mean about "implement with their role/permission system"?

Do you mean we need to implement this feature?

guzzilar: I'm talking about https://github.com/omise/omise-prestashop/pull/5#r79105766 , permission for user group, sounds nice.

nimid: @guzzilar Yes, that is right.

That page and feature has ALREDY in PrestaShop back-end. We do not need to develop anything for that permission feature.

Now the branches and pull requests are quite messy for me. I accepted that I create branches and pull requests not good.

I agree to change the type of secret key to be password type but if the changing in this point, the followed pull requests and files have affected and need to have more step in merge or rebase. That is I would like to prevent (for this situation).

If you want to do it in this point I let you to do it and solve the conflict for others pull requests.

guzzilar: ok, I got your point about branch/pr conflicted. Then I'll merge it now. thanks 👍

guzzilar (approved):

guzzilar:
👍

PR #8, opened 2016-09-13
Add a customed ruleset for PHP_CodeSniffer

nimid:
**1. Objective reason**

Ignore an error, `Each class must be in a namespace of at least one level (a top-level vendor name)`, when checking the coding standard with PHP_CodeSniffer.

To determine whether the source code meet the coding standard, it can be checked with the tool such as PHP_CodeSniffer which is faster than manually review. Including the ability to repeat the checking quickly and easily. It can reduced the time in review process and reduce the error from the human.

According to the [PrestaShop using PSR-1 and PSR-2 coding standard](http://doc.prestashop.com/display/PS16/Coding+Standards), Omise-PrestaShop has followed that standards as well.

Currently, PrestaShop stable version is 1.6. This version is not supported for namespace and from testing adding namespace will cause the system crash.

So, the rule, `PSR1.Classes.ClassDeclaration.MissingNamespace`, will be excluded when checking the coding standard with PHP_CodeSniffer to prevent an error above.

**Note:**
One way to know the rule name that PHP_CodeSniffer use for validation is to view the source code of PHP_CodeSniffer itself.

For example, from the rule above,

<pre>
<b>PSR1.Classes.ClassDeclaration</b>.MissingNamespace
</pre>

It was implemented in the class below.

<pre>
PHP_CodeSniffer/CodeSniffer/Standards/<b>PSR1/Sniffs/Classes/ClassDeclarationSniff</b>.php
</pre>

**2. Description of change**

Add a file, ruleset.xml, that contain an ignore rule.

To run PHP_CodeSniffer

`$ phpcs --standard=ruleset.xml omise/`

None

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

`-`


guzzilar:
@nimid do we need 1 break line at the end of an xml file? (saw another rulesets have).

just 1 suggestion, to avoid different local config or different version of `PHP_CodeSniffer` on a local machine, it's better to have PHP_CodeSniffer as a bundle package when we clone this repo.
(i.e. have 1 composer file that require `squizlabs/php_codesniffer` package that we could use the same version/config).

Btw, 👍 for the rest.
Do you want me to merge & rebase to #10 or you wanna do by yourself?

guzzilar (approved):
@nimid do we need 1 break line at the end of an xml file? (saw another rulesets have).

just 1 suggestion, to avoid different local config or different version of `PHP_CodeSniffer` on a local machine, it's better to have PHP_CodeSniffer as a bundle package when we clone this repo.
(i.e. have 1 composer file that require `squizlabs/php_codesniffer` package that we could use the same version/config).

Btw, 👍 for the rest.
Do you want me to merge & rebase to #10 or you wanna do by yourself?

nimid:
@guzzilar 

> do we need 1 break line at the end of an xml file? (saw another rulesets have).

Which rulesets you saw?

> have 1 composer file that require `squizlabs/php_codesniffer`

Yes, I agree.

> Do you want me to merge & rebase to #10 or you wanna do by yourself?

You can do merge and rebase.

guzzilar:
@nimid 
> Which rulesets you saw?

inside `php_codesniffer` itself.

`squizlabs/php_codesniffer/CodeSniffer/Standards/*`

i.e. `squizlabs/php_codesniffer/CodeSniffer/Standards/PSR2/ruleset.xml`

nimid:
@guzzilar Done. 0253a2cf6ae4dde239238025b0a8774d9e2f0a0c

guzzilar:
👍

PR #7, opened 2016-09-10
Display the checkout form at the front-end

nimid:
**1. Objective reason**

Display the checkout form at the front-end.

Required pull request: [#1](https://github.com/omise/omise-prestashop/pull/1), [#3](https://github.com/omise/omise-prestashop/pull/3), [#5](https://github.com/omise/omise-prestashop/pull/5) and [#6](https://github.com/omise/omise-prestashop/pull/6)

**2. Description of change**
- Implement a function, `Omise->hookPayment()`, to check module status. If the module status is valid the checkout form will be display.
- Create a checkout form template file.

The screenshot below shows the checkout form at the front-end.

screenshot-localhost 2016-09-16 03-07-02

**3. Users affected by the change**

All

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

`-`

PR #9, opened 2016-09-15
Create omise card token

nimid:
**1. Objective reason**

Create an Omise card token from the valid card information that payer filled in the checkout form. The token will be used to process the charge in the next step.

Related ticket: T1122
Required pull request: [#7](https://github.com/omise/omise-prestashop/pull/7)

**2. Description of change**
- Call the external client side script, Omise.js
- Manipulate the checkout form to create the Omise card token

The screenshot below shows the Omise card token was defined to be the value for HTML hidden after the payer filled the valid card information.

screen shot 2016-09-16 at 3 55 36 am

All

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

`-`

### DIFF
--- a/omise/form.php
+++ b/omise/form.php
@@ -1,0 +1,145 @@
+<?php
+if (! defined('_PS_VERSION_')) {
+    exit();
+}
+
+class Form extends PaymentModule
+{
+    protected $submit_action_name = 'submitOmiseForm';
+
+    public function generate()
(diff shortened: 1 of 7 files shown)


nimid:
Re-implement this feature by continue from the branch, feature/display-checkout-form.

PR #10, opened 2016-09-22
Develop the setting page by using the template file instead of using the array configuration

nimid:
**1. Objective reason**

Develop the setting page by using the template file instead of using the array configuration.

At the setting page, it has controls, secret key for test and secret key for live. For one of security best practices which is indicated on the Omise web page, [Secure your Account Secret Keys](https://www.omise.co/security-best-practices#secure-your-account-secret-keys), the type of secret key should be the password not the text.

To implement the type of control to be password and the behavior remain the same as the text, the array configuration is not appropriate. Not only the display style was changed, the core function of PrestaShop is not allowed the value to be displayed on the control that is password type when retrieve the data from the database.

So, to implement the control to be password type, using the template file is more appropriate.

An additional benefit in using the template file is fully customized.

Related ticket: T1062

**2. Description of change**
- Remove the implementation of setting page that is array configuration.
- Display the setting page by using the template file.
- Develop the function to save the setting data.
- Develop the function to retrieve the data and display it on the setting page.

The screenshot below shows the setting page that the control of secret key for test and secret key for live are the password type.

screenshot-localhost 2016-09-22 10-54-23

All

**4. Impact of the change**

`-`

**5. Priority of change**

Normal

**6. Alternate solution**

`-`

### DIFF
--- a/omise/omise.php
+++ b/omise/omise.php
@@ -28,6 +28,13 @@ class Omise extends PaymentModule
      */
     const MODULE_VERSION = '1.6.0.0';
 
+    /**
+     * The instance of class, Setting.
+     *
+     * The Setting class is used to manipulate the configuration of module.
+     *
+     * @var \Setting
+     */
     protected $setting;
 
     public function __construct()
--- a/omise/omise.php
+++ b/omise/omise.php
@@ -84,6 +84,9 @@ class Omise extends PaymentModule
         return $this->setting;
     }
 
+    /**
+     * @param \Setting $setting The instance of class, Setting.
+     */
     public function setSetting($setting)
     {
         $this->setting = $setting;
--- a/omise/omise.php
+++ b/omise/omise.php
@@ -76,6 +76,9 @@ class Omise extends PaymentModule
         return $this->display(__FILE__, 'views/templates/admin/setting.tpl');
     }
 
+    /**
+     * @return \Setting
+     */
     public function getSetting()
     {
         return $this->setting;
--- a/omise/omise.php
+++ b/omise/omise.php
@@ -15,7 +15,7 @@ class Omise extends PaymentModule
     const MODULE_DISPLAY_NAME = 'Omise';
 
     /**
-     * The name that used to referece in the program.
+     * The name that used to reference in the program.
      *
      * @var string
      */
--- a/omise/omise.php
+++ b/omise/omise.php
@@ -62,14 +62,16 @@ class Omise extends PaymentModule
             $this->smarty->assign('confirmation', $this->displayConfirmation($this->l('Settings updated')));
         }
 
-        $this->smarty->assign('live_public_key', $this->setting->getLivePublicKey());
-        $this->smarty->assign('live_secret_key', $this->setting->getLiveSecretKey());
-        $this->smarty->assign('module_status', $this->setting->isModuleEnabled());
-        $this->smarty->assign('sandbox_status', $this->setting->isSandboxEnabled());
-        $this->smarty->assign('submit_action', $this->setting->getSubmitAction());
-        $this->smarty->assign('test_public_key', $this->setting->getTestPublicKey());
-        $this->smarty->assign('test_secret_key', $this->setting->getTestSecretKey());
-        $this->smarty->assign('title', $this->setting->getTitle());
+        $this->smarty->assign(array(
+            'live_public_key' => $this->setting->getLivePublicKey(),
+            'live_secret_key' => $this->setting->getLiveSecretKey(),
+            'module_status' => $this->setting->isModuleEnabled(),
+            'sandbox_status' => $this->setting->isSandboxEnabled(),
+            'submit_action' => $this->setting->getSubmitAction(),
+            'test_public_key' => $this->setting->getTestPublicKey(),
+            'test_secret_key' => $this->setting->getTestSecretKey(),
+            'title' => $this->setting->getTitle(),
+        ));
 
         return $this->display(__FILE__, 'views/templates/admin/setting.tpl');
     }
--- a/omise/omise.php
+++ b/omise/omise.php
@@ -1,6 +1,9 @@
 <?php
-if (! defined('_PS_VERSION_'))
+if (! defined('_PS_VERSION_')) {
     exit();
+}
+
+require_once 'setting.php';
 
 class Omise extends PaymentModule
 {
@@ -25,6 +28,8 @@ class Omise extends PaymentModule
      */
     const MODULE_VERSION = '1.6.0.0';
 
+    protected $setting;
+
     public function __construct()
     {
         $this->name                   = self::MODULE_NAME;
@@ -39,101 +44,41 @@ class Omise extends PaymentModule
 
         $this->displayName            = self::MODULE_DISPLAY_NAME;
         $this->confirmUninstall       = $this->l('Are you sure you want to uninstall ' . self::MODULE_DISPLAY_NAME . ' module?');
+
+        $this->setSetting(new Setting());
     }
 
     public function getContent()
     {
-        $fields_form[0]['form'] = array(
-            'legend' => array(
-                'title' => $this->l('Settings')
-            ),
-            'input'  => array(
-                array(
-                    'type'     => 'switch',
-                    'label'    => $this->l('Enable/Disable'),
-                    'name'     => 'module',
-                    'is_bool'  => true,
-                    'desc'     => $this->l('Enable Omise Payment Module.'),
-                    'values'   => array(
-                        array(
-                            'id'    => 'module_enabled',
-                            'value' => 1,
-                            'label' => 'Enabled'
-                        ),
-                        array(
-                            'id'    => 'module_disabled',
-                            'value' => 0,
-                            'label' => 'Disabled'
-                        )
-                    )
-                ),
-                array(
-                    'type'     => 'switch',
-                    'label'    => $this->l('Sandbox'),
-                    'name'     => 'sandbox',
-                    'is_bool'  => true,
-                    'desc'     => $this->l('Enabling sandbox means that all your transactions will be in TEST mode.'),
-                    'values'   => array(
-                        array(
-                            'id'    => 'sandbox_on',
-                            'value' => 1,
-                            'label' => 'Enabled'
-                        ),
-                        array(
-                            'id'    => 'sandbox_off',
-                            'value' => 0,
-                            'label' => 'Disabled'
-                        )
-                    )
-                ),
-                array(
-                    'type'     => 'text',
-                    'label'    => $this->l('Public key for test'),
-                    'name'     => 'publicKeyForTest',
-                    'required' => false,
-                    'desc'     => 'The "Test" mode public key can be found in Omise Dashboard.'
-                ),
-                array(
-                    'type'     => 'text',
-                    'label'    => $this->l('Secret key for test'),
-                    'name'     => 'secretKeyForTest',
-                    'required' => false,
-                    'desc'     => 'The "Test" mode secret key can be found in Omise Dashboard.'
-                ),
-                array(
-                    'type'     => 'text',
-                    'label'    => $this->l('Public key for live'),
-                    'name'     => 'publicKeyForLive',
-                    'required' => false,
-                    'desc'     => 'The "Live" mode public key can be found in Omise Dashboard.'
-                ),
-                array(
-                    'type'     => 'text',
-                    'label'    => $this->l('Secret key for live'),
-                    'name'     => 'secretKeyForLive',
-                    'required' => false,
-                    'desc'     => 'The "Live" mode secret key can be found in Omise Dashboard.'
-                ),
-                array(
-                    'label'    => '<b>Advance Settings</b>'
-                ),
-                array(
-                    'type'     => 'text',
-                    'label'    => $this->l('Title'),
-                    'name'     => 'title',
-                    'required' => false,
-                    'desc'     => 'This controls the title which the user sees during checkout.'
-                )
-            ),
-            'submit' => array(
-                'title' => $this->l('Save'),
-                'class' => 'btn btn-default pull-right'
-            )
-        );
+        if ($this->setting->isSubmit()) {
+            $this->setting->save();
+            $this->smarty->assign('confirmation', $this->displayConfirmation($this->l('Settings updated')));
+        }
 
-        $helper = new HelperForm();
-        $helper->submit_action = 'submit' . $this->name;
+        $this->smarty->assign('live_public_key', $this->setting->getLivePublicKey());
+        $this->smarty->assign('live_secret_key', $this->setting->getLiveSecretKey());
+        $this->smarty->assign('module_status', $this->setting->isModuleEnabled());
+        $this->smarty->assign('sandbox_status', $this->setting->isSandboxEnabled());
+        $this->smarty->assign('submit_action', $this->setting->getSubmitAction());
+        $this->smarty->assign('test_public_key', $this->setting->getTestPublicKey());
+        $this->smarty->assign('test_secret_key', $this->setting->getTestSecretKey());
+        $this->smarty->assign('title', $this->setting->getTitle());
 
-        return $helper->generateForm($fields_form);
+        return $this->display(__FILE__, 'views/templates/admin/setting.tpl');
+    }
+
+    public function getSetting()
+    {
+        return $this->setting;
+    }
+
+    public function setSetting($setting)
+    {
+        $this->setting = $setting;
+    }
+
+    public function setSmarty($smarty)
+    {
+        $this->smarty = $smarty;
     }
 }
--- a/omise/omise.php
+++ b/omise/omise.php
@@ -92,6 +92,9 @@ class Omise extends PaymentModule
         $this->setting = $setting;
     }
 
+    /**
+     * @param \Smarty_Data $smarty The instance of class, Smarty_Data.
+     */
     public function setSmarty($smarty)
     {
         $this->smarty = $smarty;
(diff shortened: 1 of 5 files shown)


omise/omise.php
```diff
@@ -25,6 +28,8 @@ class Omise extends PaymentModule
      */
     const MODULE_VERSION = '1.6.0.0';
 
+    protected $setting;
```
guzzilar, line 38: @nimid can you add docblock for this

tests/unit/OmiseTest.php
```diff
@@ -1,37 +1,121 @@
...
+        $this->omise->setSmarty($this->smarty);
+    }
+
+    public function testConstructor_whenNewTheInstance_theDefaultValueOfTheAttributeSettingMustBeAvailable()
+    {
+        $omise = new Omise();
+
+        $setting = $omise->getSetting();
+
+        $this->assertNotEmpty($setting);
```
guzzilar: @nimid maybe better to assert if `$setting` is an instance of `Setting` class.

tests/unit/OmiseTest.php
```diff
@@ -1,37 +1,121 @@
...
+                )
+            )
             ->getMock();
 
         $this->omise = new Omise();
+        $this->omise->setSetting($this->setting);
+        $this->omise->setSmarty($this->smarty);
+    }
+
+    public function testConstructor_whenNewTheInstance_theDefaultValueOfTheAttributeSettingMustBeAvailable()
```
guzzilar: @nimid `whenNewTheInstance` can rewrite to `whenInitiateTheNewInstance` or `whenCreateTheNewInstance`

omise/setting.php
```diff
@@ -0,0 +1,86 @@
...
+        return Configuration::get('live_secret_key');
+    }
+
+    public function getPublicKey()
+    {
+        if ($this->isSandboxEnabled()) {
+            return Configuration::get('test_public_key');
+        }
+
+        return Configuration::get('live_public_key');
```
guzzilar, line 32: @nimid I was commenting to use `$this->getTestPublicKey()`, `$this->getLivePublicKey()` instead of use `Configuration::get()` but saw you already changed it in the next PR (#11) https://github.com/omise/omise-prestashop/blob/feature/display-checkout-form/omise/setting.php#L29

Shall I skip review it from this PR? or maybe we can move code from that PR to this PR?
(Im ok both)

omise/setting.php
```diff
@@ -0,0 +1,86 @@
+<?php
+if (! defined('_PS_VERSION_')) {
+    exit();
+}
+
+class Setting extends PaymentModule
+{
+    protected $submit_action = 'omise_save_setting';
```
guzzilar, line 8: @nimid Can it be `const SUBMIT_ACTION`?

omise/setting.php
```diff
@@ -0,0 +1,86 @@
+<?php
```
guzzilar, line 1: @nimid pls add docblock to all of these methods?

tests/unit/OmiseTest.php
```diff
@@ -1,37 +1,121 @@
...
+                array('submit_action', 'submit_action'),
+                array('test_public_key', 'test_public_key'),
+                array('test_secret_key', 'test_secret_key'),
+                array('title', 'title')
+            );
+
+        $this->omise->getContent();
+    }
+
+    public function testGetContent_merchantSaveSetting_theSettingDataMustBeSaved()
```
guzzilar: @nimid this test assert that method `save` will be triggered if `isSubmit()` return `true`.
might be better to rewrite the test name to 

"testGetContent_merchantSaveSetting_~theSettingDataMustBeSaved~**theSettingSaveMethodMustBeTriggered**" 

to make it more clear.

tests/unit/OmiseTest.php
```diff
@@ -1,37 +1,121 @@
...
     }
 
-    public function testDisplayName_Omise()
+    public function testDisplayName_theNameThatUsedToDisplayeToTheMerchantMustBe_Omise()
     {
         $this->assertEquals('Omise', $this->omise->displayName);
     }
 
-    public function testNeedInstance_0()
+    public function testNeedInstance_noNeedToLoadModuleAtTheBackendModulePage()
```
guzzilar: @nimid this test's missing "the expectation" from its name, is it better to verbose more?
i.e. `testNeedInstance_noNeedToLoadModuleAtTheBackendModulePageAndReturn_0` ?
or
`testNeedInstanceAttribute_willReturn_0`.

tests/unit/OmiseTest.php
```diff
@@ -1,37 +1,121 @@
...
     {
         $this->assertEquals(0, $this->omise->need_instance);
     }
 
-    public function testBootstrap_true()
+    public function testBootstrap_bootstrapTemplateIsRequired()
```
guzzilar: @nimid same as https://github.com/omise/omise-prestashop/pull/10/files#r95522979

guzzilar: btw, Im ok to skip that part if that "the expectation" is an optional, from
https://github.com/omise/omise-prestashop/pull/3#pullrequestreview-94529

guzzilar (changes requested):

omise/setting.php
```diff
@@ -0,0 +1,86 @@
...
+if (! defined('_PS_VERSION_')) {
+    exit();
+}
+
+class Setting extends PaymentModule
```
guzzilar: @nimid do we need `PaymentModule` here? (I tried remove and it still works)

omise/omise.php
```diff
@@ -39,101 +44,41 @@ public function __construct()
...
-        $helper = new HelperForm();
-        $helper->submit_action = 'submit' . $this->name;
+        $this->smarty->assign('live_public_key', $this->setting->getLivePublicKey());
+        $this->smarty->assign('live_secret_key', $this->setting->getLiveSecretKey());
+        $this->smarty->assign('module_status', $this->setting->isModuleEnabled());
+        $this->smarty->assign('sandbox_status', $this->setting->isSandboxEnabled());
+        $this->smarty->assign('submit_action', $this->setting->getSubmitAction());
+        $this->smarty->assign('test_public_key', $this->setting->getTestPublicKey());
+        $this->smarty->assign('test_secret_key', $this->setting->getTestSecretKey());
+        $this->smarty->assign('title', $this->setting->getTitle());
```
guzzilar: @nimid fyi, `$this->smarty->assign();` could assign array. 
i.e.
```php
$this->smarty->assign(array(
    'test_public_key' => $this->setting->getTestPublicKey(),
    'live_public_key' => $this->setting->getLivePublicKey(),
));
```

omise/omise.php
nimid: @guzzilar Done. 02db00d9f417195ad08da4b54e9499ec25fbdd9b

tests/unit/OmiseTest.php
nimid: @guzzilar Done. ec575f475885ddbbf8fea8bc0db07d1c4a295e9c

tests/unit/OmiseTest.php
nimid: @guzzilar Done. 726d2cd634e753ad7f9942dd27e45b1de669a553

tests/unit/OmiseTest.php
nimid: @guzzilar Done. 9baab44bfca339ab041f7ab239ab9f5b72e59100

tests/unit/OmiseTest.php
nimid: @guzzilar Done. c1273c1a71e056af1cff615b0f06db45a5242eb1

omise/setting.php
nimid: @guzzilar Done. 7797ced697e0a40d00ea811e34afa7d5335f23e4

A function, getPublicKey(), has been added the DocBlock in the [next pull request](https://github.com/omise/omise-prestashop/pull/11/files#diff-726b73960d58daf9fe48ddd631df8a5cR21).

omise/setting.php
nimid: @guzzilar Can you provide the objective and advantage?

omise/setting.php
nimid: @guzzilar It has been improved in the next pull request. So, it is not necessary to comment here.

omise/setting.php
nimid: @guzzilar Done. b3cfa66de0e52e5d8b192caccd521480d6480ace

omise/setting.php
guzzilar: @nimid well, for this case just `const` make code more understandable that this field will not be changed by any part of code.
Btw, can you share the idea behind that you chose `protected $submit_action`?

omise/setting.php
guzzilar: 👌

omise/setting.php
guzzilar: 👌

tests/unit/OmiseTest.php
nimid: @guzzilar The current expectation part, `theSettingDataMustBeSaved`, is the abstraction level. It has been abstracted from the implementation.

From an example, `theSettingSaveMethodMustBeTriggered`, if the implementation has been changed from `$setting->save()` to `$configuration->update()`, the expectation part need to be changed to conform with the implementation. That is double change (the implementation and the test case name).

To make more example, with the pattern from an example applied to the condition or situation part of the test case name, the current situation part is `merchantSaveSetting`. It might be changed to `submitIsTrue` or others. From that change, it is not guide to any part of how program, process or step work.

But with the abstraction level, the implementation can be changed to anything. The expectation of the unit is still the same, `theSettingDataMustBeSaved`.

But I did not say everything or every test case name can be or must be abstracted. It still need to be considered.

1. How about the beyond benefit of the unit test case name can capture, guide or imply the business logic or requirement objective?

2. Can you provide another example of the expectation part?

guzzilar: @nimid understand your point.

But it's not wrong to double change (the implementation part & test case name).
Since this's unit test, so you're going to test the smallest unit of functionality which is you expected that the `save()` method from `Setting` will be triggered once and that's it, right?

Well, let me say in this direction, if that implementation code be triggered once by `Omise::getContent()` properly but the `Setting::save()` method can't really save the data to the DB (maybe, somehow we forget to execute `Configuration::updateValue()` function)

So, the statement `theSettingDataMustBeSaved` will be false but the test will pass.

Instead, your test name will be pretty perfect for `integration test` that we can really test save the data to DB and assert by retrieving it back. And that time we could say `testGetContent_merchantSaveSetting_theSettingDataMustBeSaved`.

omise/setting.php
```diff
@@ -0,0 +1,113 @@
+<?php
+if (! defined('_PS_VERSION_')) {
+    exit();
+}
+
+class Setting
+{
+    protected $submit_action = 'omise_save_setting';
```
guzzilar, line 8: @nimid continue from https://github.com/omise/omise-prestashop/pull/10#discussion_r95589021 (not sure why the comment doesn't show here 🤔 )

I just found that if we use `const SUBMIT_ACTION = 'omise_save_setting';`
we can get rid of `Setting::getSubmitAction()` method from everywhere (at omise.php & OmiseTest.php).

tests/unit/OmiseTest.php
guzzilar: @nimid how about this one
`testGetContent_merchantSaveSetting_mustTriggerAProperMethodToSaveTheData`

more abstract but still related with the test.

guzzilar:
@nimid just only 2 main comments left:
- https://github.com/omise/omise-prestashop/pull/10/files#r95669131
- https://github.com/omise/omise-prestashop/pull/10/files#r95669803

and ok for the rest 👍

guzzilar:
fyi, there still has some methods that docblock is missing.

guzzilar:
fyi, there still has some methods that docblock is missing.
Here:
- https://github.com/omise/omise-prestashop/pull/10/files#diff-9b969693cd451a46a870fab6297af91dR77

- https://github.com/omise/omise-prestashop/pull/10/files#diff-9b969693cd451a46a870fab6297af91dR82

- https://github.com/omise/omise-prestashop/pull/10/files#diff-9b969693cd451a46a870fab6297af91dR87

- https://github.com/omise/omise-prestashop/pull/10/files#diff-726b73960d58daf9fe48ddd631df8a5cR103

thanks! 👍

guzzilar:
@nimid ✋ do you have any rule behind that you named these filename
`setting.php`, `checkout_form.php` (from #11) instead of `Setting.php`, `CheckoutForm.php`?

saw from test classes you used `OmiseTest.php`, `CheckoutFormTest.php`.

tests/unit/OmiseTest.php
nimid: @guzzilar 

> But it's not wrong to double change

It is double developers effort.

> which is you expected that the save() method from Setting will be triggered once and that's it, right?

No, I expect the setting data must be saved.

> but the Setting::save() method can't really save the data to the DB (maybe, somehow we forget to execute Configuration::updateValue() function)

It is out of scope of a class, `Omise`. The scope of a unit test case `OmiseTest` is a class `Omise`. It does not need to care another class or dependency.

Each unit test class will test a base class and ensure the base class is correctly work.

> So, the statement theSettingDataMustBeSaved will be false but the test will pass.

With unit test case it do not really know and do not care that another class will pass or fail because all dependencies have been mocked.

---

I still prefer a benefit of unit test case that can has more value in part of description more than just for test a method has been executed.

This benefit can be given to other developers or contributors. Who do not have the knowledge or have the little knowledge about the system to be able more understand.

For this time, you can change the expectation part to what you understand and I will review it.

omise/setting.php
nimid: @guzzilar 

> for this case just const make code more understandable that this field will not be changed by any part of code.

Yes, everything has pros and cons. With class constant, it can not be changed and extended.

> can you share the idea behind that you chose protected $submit_action?

The idea is extendability.

omise/omise.php
nimid: @guzzilar Done. 46862323ec3ee2ee7b3d89d61e95819c93cd009b

nimid:
@guzzilar For a function, `Setting.save()` https://github.com/omise/omise-prestashop/pull/10/files#diff-726b73960d58daf9fe48ddd631df8a5cR103`, which DocBlock tag you considered it is missing.

nimid:
@guzzilar 
> do you have any rule behind that you named these filename

At a table that in the section, **File structure for a PrestaShop module**. http://doc.prestashop.com/display/PS16/Creating+a+first+module

guzzilar:
@nimid https://github.com/omise/omise-prestashop/pull/10#issuecomment-272377747
>  For a function, Setting.save() ... which DocBlock tag you considered it is missing.

thought we can add
```php
/**
 * @return void
 */
```

omise/setting.php
guzzilar: @nimid ✋ more question.
>  it can not be changed and extended.

Why it need to be able to be changed? (not sure which case that we will change it.)
Also, which case that it need to be able to be extended?

guzzilar:
@nimid https://github.com/omise/omise-prestashop/pull/10#issuecomment-272378189

> At a table that in the section, File structure for a PrestaShop module.

checked it, can't find the right answer there 😓 (saw only they mentioned about only one file, `name_of_the_module.php`)

Guess you want to make it corresponding with `name_of_the_module.php` right?

_Btw, just because usually I saw filename that name it as `MyClass.php` to make it corresponding to its class name `class MyClass`. 
But, anyway also saw some PrestaShop's modules named their files in various ways.
(some of them are, `Do_Something.php`, `DoSomething.php` or `do_something.php`)_

_just wanna understand the idea behind :)_

tests/unit/OmiseTest.php
guzzilar: > For this time, you can change the expectation part to what you understand and I will review it.

What  is a point thou 😂 

> I still prefer a benefit of unit test case that can has more value in part of description more than just for test a method has been executed.

Ok, I just feel strange with mustBe`Saved`  but it's not `saved`.

So when I read this test I expect that this test will prove that the data really do "save" and that's it.

Still, understand your point :)

btw, to make it short, do you like `testGetContent_merchantSaveSetting_mustTriggerAProperMethodToSaveTheData`?

if not, just go with this.

nimid:
@guzzilar 

I agree with `@return void` but I have researched and found [a description](https://phpdoc.org/docs/latest/references/phpdoc/tags/return.html#description).

> 2. functions and methods without a return value, the `@return` tag MAY be omitted here, in which case `@return` void is implied.

guzzilar:
https://github.com/omise/omise-prestashop/pull/10#issuecomment-272390907
@nimid I think that will work for `__constructor()` that we usually skip `@return`.

Anyway, 👌

omise/setting.php
nimid: @guzzilar Sorry, I mean our customers or users.

guzzilar: @nimid 👌 got it.

tests/unit/OmiseTest.php
nimid: @guzzilar No, I do not like.

How about `testGetContent_merchantSaveSetting_saveTheSettingData`?

guzzilar: @nimid better 👍

nimid: @guzzilar Done. f1f433a73f3e544a5482b80730cf3d285bf4c450

guzzilar (approved):

guzzilar:
👍 👍 👍

PR #18, opened 2017-01-09
Implement function to save an order after created Omise charge

nimid:
#### 1. Objective

Implement function to save an order after created Omise charge.

**Related information**:
Related issue: -
Related ticket: T1147
Required pull request: #19 

#### 2. Description of change

- Add a new class, `PaymentOrder`, to save the order
- Implement a function that used to redirect the system to the result page
- Modify payment controller class, `OmisePaymentModuleFrontController`, to save an order and redirect to the result page

#### 3. Quality assurance

**Environments:**

- **Platform**: PrestaShop 1.6.1.6
- **Omise plugin**: Omise-PrestaShop 1.6.0.0
- **PHP**: 5.6.28
- **Mozilla Firefox**: 50.1.0

**Details:**

- Make sure that the Omise-PrestaShop has been installed and enabled.
- At the front-end, add a product to cart.
- Proceed to checkout.
- At the step, 05. Payment, fill the valid card information and click button, Submit Payment. The system will process the payment and display the result page.

The screenshot below shows a product that will be used to test.

create_order

The screenshot below shows the default result page of PrestaShop. After successfully created Omise charge and saved an order, the cart information will be cleared.

cart_empty

The screenshot below shows the saved order from the above payment at the _payer_ side.

success_save_customer_side

The screenshot below shows the saved order from the above payment at the _merchant_ side.

success_save_merchant_side

#### 4. Impact of the change

`-`

#### 5. Priority of change

Normal

#### 6. Additional notes

The result page still be the default of PrestaShop. It has no any information about the order. For the next pull request, the proper result page will be developed including error handling at the server side.

The screenshot below shows the [passed unit tests at Travis CI](https://travis-ci.org/nimid/omise-prestashop/builds/190196278).

screenshot-travis-ci org 2017-01-09 16-54-45

### DIFF
--- a/omise/classes/payment_order.php
+++ b/omise/classes/payment_order.php
@@ -46,6 +46,16 @@ class PaymentOrder extends ModuleFrontController
         return $this->module->displayName;
     }
 
+    protected function getModuleId()
+    {
+        return (int) $this->module->id;
+    }
+
+    protected function getModuleCurrentOrder()
+    {
+        return $this->module->currentOrder;
+    }
+
     /**
      * The optional message that will be used to attach to the order.
      *
@@ -78,6 +88,16 @@ class PaymentOrder extends ModuleFrontController
     protected function isNotNeededRoundingCardOrderTotal()
     {
         return false;
+    }
+
+    public function redirectToResultPage()
+    {
+        Tools::redirect('index.php?controller=order-confirmation' .
+            '&id_cart=' . $this->getCartId() .
+            '&id_module=' . $this->getModuleId() .
+            '&id_order=' . $this->getModuleCurrentOrder() .
+            '&key=' . $this->getCustomerSecureKey()
+        );
     }
 
     public function save()
--- a/omise/classes/payment_order.php
+++ b/omise/classes/payment_order.php
@@ -3,6 +3,6 @@ if (! defined('_PS_VERSION_')) {
     exit();
 }
 
-class PaymentOrder
+class PaymentOrder extends ModuleFrontController
 {
 }
--- a/omise/classes/payment_order.php
+++ b/omise/classes/payment_order.php
@@ -5,4 +5,93 @@ if (! defined('_PS_VERSION_')) {
 
 class PaymentOrder extends ModuleFrontController
 {
+    protected function getCartId()
+    {
+        $cart = $this->context->cart;
+        return (int) $cart->id;
+    }
+
+    protected function getCartOrderTotal()
+    {
+        $cart = $this->context->cart;
+        return (float) $cart->getOrderTotal();
+    }
+
+    protected function getCustomerSecureKey()
+    {
+        $cart = $this->context->cart;
+        $customer = new Customer($cart->id_customer);
+
+        return $customer->secure_key;
+    }
+
+    protected function getCurrencyId()
+    {
+        $currency = $this->context->currency;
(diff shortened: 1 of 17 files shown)


omise/controllers/front/payment.php
```diff
@@ -0,0 +1,43 @@
...
+class OmisePaymentModuleFrontController extends ModuleFrontController
+{
+    public $display_column_left = false;
+    protected $payment_order;
+
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->setPaymentOrder(new PaymentOrder());
```
guzzilar, line 20: @nimid ✋ have question
At the `initContent()` method you create new `Charge` instance there but for `PaymentOrder` instance you chose to assign to a class's property here.

Why not be something like,
```php
public function __construct()
{
    $this->setPaymentOrder(new PaymentOrder())
    $this->setCharge(new Charge())
}
```

or
```php
public function initContent()
{
    $paymentOrder = new PaymentOrder();
    $omiseCharge = new Charge();
}
```

omise/classes/payment_order.php
```diff
@@ -0,0 +1,117 @@
+<?php
+if (! defined('_PS_VERSION_')) {
+    exit();
+}
+
+class PaymentOrder extends ModuleFrontController
```
guzzilar, line 6: @nimid If classes/payment_order.php is a controller, I think it better to keep it in controllers folder. (saw that this class extends a parent class from `ModuleFrontController`).

omise/classes/payment_order.php
```diff
@@ -0,0 +1,117 @@
...
+{
+    protected function getCartId()
```
guzzilar, line 8: @nimid how about just have method 

```php
protected function cart()
{
    return $this->context->cart;
}
```

Then you just call 
```php
$this->cart()->id;
$this->cart()->id_customer;
$this->cart()->getOrderTotal();
```

or even you can call `$this->context->cart->id` directly without providing getter methods.

omise/classes/payment_order.php
```diff
@@ -0,0 +1,117 @@
...
+    {
+        $cart = $this->context->cart;
+        return (int) $cart->id;
```
guzzilar, line 11: @nimid can we just `return (int) $this->context->cart->id;` ?

omise/classes/payment_order.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    protected function getCartOrderTotal()
+    {
+        $cart = $this->context->cart;
+        return (float) $cart->getOrderTotal();
```
guzzilar, line 17: @nimid can we just `return (float) $this->context->cart->getOrderTotal();`?

omise/classes/payment_order.php
```diff
@@ -0,0 +1,117 @@
...
+        $cart = $this->context->cart;
+        $customer = new Customer($cart->id_customer);
+
+        return $customer->secure_key;
+    }
+
+    protected function getCurrencyId()
+    {
+        $currency = $this->context->currency;
+        return (int) $currency->id;
```
guzzilar, line 31: @nimid same as above, how about `return (int) $this->context->currency->id;` ?

nimid:
The source code in this pull request has been changed and combined to merge together with the pull request #21.

PR #17, opened 2017-01-07
Add a unit test for a controller class, OmisePaymentModuleFrontController

nimid:
#### 1. Objective

- Add a unit test for a controller class, `OmisePaymentModuleFrontController`.
- Emphasize the dependencies of a controller class, `OmisePaymentModuleFrontController `. The dependencies should be reduced.

**Related information**:
Related issue: -
Related ticket: -
Required pull request: #14 

#### 2. Description of change

- Add a dependency, Mockery, for the unit testing. This dependency has been used to create mock object for the static methods.
- Using the Composer to be the dependency management.
- Add a unit test for a class, `OmisePaymentModuleFrontController`.
- Un-track a directory, vendor, from Git by add a new entry to the list of gitignore. This directory has been used to store the local dependency which it does not need to store in the repository.

#### 3. Quality assurance

**Environments:**

- **Platform version**: - (This pull request did not change the base source code. Therefore, it is not required to test with platform.)
- **Omise plugin version**: Omise-PrestaShop 1.6.0.0
- **PHP versions**: 5.4, 5.5, 5.6, 7.0 and 7.1

**Details:**

**Manual testing**
Run the unit testing with PHPUnit (no additional required parameters).
For example:

```
$ phpunit
```

**Automated testing**
Setup the system such as Travis CI, CircleCI or Jenkins.

#### 4. Impact of the change

`-`

#### 5. Priority of change

Normal

#### 6. Additional notes

- There are 2 methods have not been tested.
  1. The method `getAmount()` has not been tested because it has a [reference to another class constant, `Cart::BOTH`](https://github.com/omise/omise-prestashop/blob/feature/refactor-extract-payment-controller-class/omise/controllers/front/payment.php#L41). This reference is not easy to mock with PHPUnit or Mockery. Although, to test this method, it can be used another technique by create stub but for this time, create stub has more expensive.
  2. The method `initContent()` has not been tested too because this method depends on the above method, `getAmount()`.

The screenshot below shows the [passed unit tests at Travis CI](https://travis-ci.org/nimid/omise-prestashop/builds/189119095).

pass_at_travis_ci

The screenshot below shows the [passed unit tests at CircleCI](https://circleci.com/gh/nimid/omise-prestashop/41#tests/containers/0).

pass_at_circleci

PR #16, opened 2017-01-05
Refactor the process of create Omise charge by extract class

nimid:
#### 1. Objective

Reduce the dependencies and functionality of a controller class, `OmisePaymentModuleFrontController`, by using a refactoring technique, [extract class](https://refactoring.com/catalog/extractClass.html).

The `OmisePaymentModuleFrontController` has been used to create Omise charge.

According to the pull request #17, it has emphasized the class, `OmisePaymentModuleFrontController `, has many dependencies and functionality. This situation may tend to the class has error prone, especially in part of software maintenance.

Including the next steps, this class will been modified to add more functions such as store the order, handle the errors and display the proper result.

So, it will be better to reduce the dependencies and functionality of a class, `OmisePaymentModuleFrontController`.

**Related information**:
Related issue: -
Related ticket: -
Required pull request: #17 

#### 2. Description of change

- Add a new delegate class, `Charge`, to create Omise charge request and interface with Omise-PHP.
- Remove all functions in controller class, `OmisePaymentModuleFrontController`, that have been moved to delegate class.

#### 3. Quality assurance

**Environments:**

- **Platform**: PrestaShop 1.6.1.6
- **Omise plugin**: Omise-PrestaShop 1.6.0.0
- **PHP**: 5.6.28
- **Mozilla Firefox**: 50.1.0

**Details:**

- Make sure that the Omise-PrestaShop has been installed and enabled.
- At the front-end, add a product to cart.
- Proceed to checkout.

**Test case 1: Successfully checkout**
- At the step, 05. Payment, fill the valid card information and click button, Submit Payment. The system will process the payment and display the result page.

The screenshot below shows the result page display the successful payment.

prestashop-checkout-success

The screenshot below shows charge detail at Omise dashboard. This charge was created from the above checkout.

omise-dashboard-success-charge

**Test case 2: Failed checkout**
- From the above success checkout, refresh the web browser page.

The screenshot below shows the result page display the error payment.

prestashop-checkout-fail

#### 4. Impact of the change

`-`

#### 5. Priority of change

Normal

#### 6. Additional notes

The function to handle the create Omise charge error will be developed for the next pull request.

The screenshot below shows the [passed unit tests at Travis CI](https://travis-ci.org/nimid/omise-prestashop/builds/189773008).

travis-success

PR #14, opened 2016-12-26
Create an Omise charge

nimid:
#### 1. Objective

Create an Omise charge by using the Omise card token that submitted from the client side.

**Related information**:
Related issue: -
Related ticket: -
Required pull request: #13 

#### 2. Description of change

- Add a helper class and a function, `OmisePluginHelperCharge.amount()` to format the charge amount for the Thai Baht currency.
- Complete the checkout form function to submit the Omise card token from client side to server side.
- Implement a class to process charge at server side.
- Add a dummy page for display the checkout result.

#### 3. Quality assurance

**Environments:**

- **Platform**: PrestaShop 1.6.1.6
- **Omise plugin**: Omise-PrestaShop 1.6.0.0
- **PHP**: 5.6.28
- **Mozilla Firefox**: 50.1.0

**Details:**

- Make sure that the Omise-PrestaShop has been installed and enabled.
- At the front-end, add a product to cart.
- Proceed to checkout.
- At the step, 05. Payment, fill the valid card information and click button, Submit Payment. The system will process the payment and display the result page.

The screenshot below shows the result page display the successful payment.

screenshot-localhost 1616 2016-12-26 14-19-39

The screenshot below shows charge detail at Omise dashboard. This charge was created from the above checkout.

screenshot-dashboard omise co 2016-12-26 14-21-32

#### 4. Impact of the change

`-`

#### 5. Priority of change

Normal

#### 6. Additional notes

For the next steps of charge process, such as record the order information to PrestaShop or display the proper result page, it will be developed for next pull request.

PR #13, opened 2016-12-25
Bundle Omise-PHP to the project

nimid:
#### 1. Objective

Bundle a library, Omise-PHP, to the project. Omise-PHP will be used to interface with Omise API.

**Related information**:
Related issue: -
Related ticket: -
Required pull request: #12 

#### 2. Description of change

- Add Omise-PHP to the project in the directory, `libraries\omise-php`
- A directory and files of Omise-PHP in the list below are not added to the project.
  1 `tests\`
  2 `.gitignore`
  3 `circle.yml`
  4 `composer.json`
  5 `phpunit.xml`

#### 3. Quality assurance

**Environments:**

- **Platform**: PrestaShop 1.6.1.6
- **Omise plugin**: Omise-PrestaShop 1.6.0.0
- **PHP**: 5.6.28
- **Omise-PHP**: 2.6.0

**Details:**

This pull request does not required testing.

#### 4. Impact of the change

`-`

#### 5. Priority of change

Normal

#### 6. Additional Notes

`-`

PR #19, opened 2017-01-18
Create an Omise charge

guzzilar:
> Note1, original PR was made by @nimid at PR #14, #17, #16.
> This PR just did (and aimed to) rearrange those 3 PRs to make it easy to review by merging all those 3 PRs to only one, which is, this branch.
>
> Note2, this branch contains the same code at PR #16 (except, this branch already rebased with `develop`.

#### 1. Objective

Create an Omise charge by using the Omise card token that submitted from the client side.

**Related information**:
Related issue: -
Related ticket: -
Required pull request: #13 

#### 2. Description of change

- Add a helper class and a function, `OmisePluginHelperCharge.amount()` to format the charge amount for the Thai Baht currency.
- Complete the checkout form function to submit the Omise card token from client side to server side.
- Implement a class to process charge at server side.
- Add a dummy page for display the checkout result.

#### 3. Quality assurance

**Environments:**

- **Platform**: PrestaShop 1.6.1.6
- **Omise plugin**: Omise-PrestaShop 1.6.0.0
- **PHP**: 5.6.28
- **Mozilla Firefox**: 50.1.0

**Details:**

- Make sure that the Omise-PrestaShop has been installed and enabled.
- At the front-end, add a product to cart.
- Proceed to checkout.
- At the step, 05. Payment, fill the valid card information and click button, Submit Payment. The system will process the payment and display the result page.

The screenshot below shows the result page display the successful payment.

screenshot-localhost 1616 2016-12-26 14-19-39

The screenshot below shows charge detail at Omise dashboard. This charge was created from the above checkout.

screenshot-dashboard omise co 2016-12-26 14-21-32

#### 4. Impact of the change

`-`

#### 5. Priority of change

Normal

#### 6. Additional notes

For the next steps of charge process, such as record the order information to PrestaShop or display the proper result page, it will be developed for next pull request.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,5 @@
+{
+    "require-dev": {
+        "mockery/mockery": "dev-master"
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,9 @@
 {
     "require-dev": {
         "mockery/mockery": "dev-master"
+    },
+    "autoload": {
+        "classmap": ["omise/"],
+        "exclude-from-classmap": ["/omise/libraries/"]
     }
 }
--- a/omise/classes/charge.php
+++ b/omise/classes/charge.php
@@ -1,0 +1,9 @@
+<?php
+if (! defined('_PS_VERSION_')) {
+    exit();
+}
+
+class Charge
+{
+
+}
--- a/omise/classes/charge.php
+++ b/omise/classes/charge.php
@@ -16,7 +16,7 @@ class Charge extends ModuleFrontController
         $charge_request = array(
             'amount' => $this->getAmount(),
             'card' => $this->getCardToken(),
-            'capture' => $this->getCapture(),
+            'capture' => 'true',
             'currency' => $this->getCurrencyCode(),
             'description' => $this->getChargeDescription(),
         );
@@ -35,11 +35,6 @@ class Charge extends ModuleFrontController
     protected function getCardToken()
     {
         return Tools::getValue('omise_card_token');
-    }
-
-    protected function getCapture()
-    {
-        return 'true';
     }
 
     protected function getChargeDescription()
--- a/omise/classes/charge.php
+++ b/omise/classes/charge.php
@@ -3,7 +3,47 @@ if (! defined('_PS_VERSION_')) {
     exit();
 }
 
+require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-php/lib/Omise.php';
+require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-plugin/Omise.php';
+require_once _PS_MODULE_DIR_ . '/omise/setting.php';
+
 class Charge extends ModuleFrontController
 {
+    public function getAmount()
+    {
+        $currency_code = $this->getCurrencyCode();
+        $order_total = (float) $this->context->cart->getOrderTotal(true, Cart::BOTH);
 
+        return OmisePluginHelperCharge::amount($currency_code, $order_total);
+    }
+
+    public function getCardToken()
+    {
+        return Tools::getValue('omise_card_token');
+    }
+
+    public function getCapture()
+    {
+        return 'true';
+    }
+
+    public function getChargeDescription()
+    {
+        return 'Charge a card using a token from PrestaShop (' . _PS_VERSION_ . ')';
+    }
+
+    public function getCurrencyCode()
+    {
+        $currency_id = (int) $this->context->cart->id_currency;
+        $currency_instance = Currency::getCurrencyInstance($currency_id);
+
+        return $currency_instance->iso_code;
+    }
+
+    public function getSecretKey()
+    {
+        $setting = new Setting();
+
+        return $setting->getSecretKey();
+    }
 }
--- a/omise/classes/charge.php
+++ b/omise/classes/charge.php
@@ -3,9 +3,11 @@ if (! defined('_PS_VERSION_')) {
     exit();
 }
 
-require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-php/lib/Omise.php';
-require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-plugin/Omise.php';
-require_once _PS_MODULE_DIR_ . '/omise/setting.php';
+if (defined('_PS_MODULE_DIR_')) {
+    require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-php/lib/Omise.php';
+    require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-plugin/Omise.php';
+    require_once _PS_MODULE_DIR_ . '/omise/setting.php';
+}
 
 class Charge extends ModuleFrontController
 {
--- a/omise/classes/charge.php
+++ b/omise/classes/charge.php
@@ -9,7 +9,7 @@ require_once _PS_MODULE_DIR_ . '/omise/setting.php';
 
 class Charge extends ModuleFrontController
 {
-    public function getAmount()
+    protected function getAmount()
     {
         $currency_code = $this->getCurrencyCode();
         $order_total = (float) $this->context->cart->getOrderTotal(true, Cart::BOTH);
@@ -17,22 +17,22 @@ class Charge extends ModuleFrontController
         return OmisePluginHelperCharge::amount($currency_code, $order_total);
     }
 
-    public function getCardToken()
+    protected function getCardToken()
     {
         return Tools::getValue('omise_card_token');
     }
 
-    public function getCapture()
+    protected function getCapture()
     {
         return 'true';
     }
 
-    public function getChargeDescription()
+    protected function getChargeDescription()
     {
         return 'Charge a card using a token from PrestaShop (' . _PS_VERSION_ . ')';
     }
 
-    public function getCurrencyCode()
+    protected function getCurrencyCode()
     {
         $currency_id = (int) $this->context->cart->id_currency;
         $currency_instance = Currency::getCurrencyInstance($currency_id);
@@ -40,7 +40,7 @@ class Charge extends ModuleFrontController
         return $currency_instance->iso_code;
     }
 
-    public function getSecretKey()
+    protected function getSecretKey()
     {
         $setting = new Setting();
 
--- a/omise/classes/charge.php
+++ b/omise/classes/charge.php
@@ -9,6 +9,19 @@ require_once _PS_MODULE_DIR_ . '/omise/setting.php';
 
 class Charge extends ModuleFrontController
 {
+    public function create()
+    {
+        $charge_request = array(
+            'amount' => $this->getAmount(),
+            'card' => $this->getCardToken(),
+            'capture' => $this->getCapture(),
+            'currency' => $this->getCurrencyCode(),
+            'description' => $this->getChargeDescription(),
+        );
+
+        return OmiseCharge::create($charge_request, '', $this->getSecretKey());
+    }
+
     protected function getAmount()
     {
         $currency_code = $this->getCurrencyCode();
--- a/omise/classes/charge.php
+++ b/omise/classes/charge.php
@@ -3,7 +3,7 @@ if (! defined('_PS_VERSION_')) {
     exit();
 }
 
-class Charge
+class Charge extends ModuleFrontController
 {
 
 }
--- a/omise/controllers/front/payment.php
+++ b/omise/controllers/front/payment.php
@@ -14,18 +14,10 @@ class OmisePaymentModuleFrontController extends ModuleFrontController
     {
         parent::initContent();
 
-        $charge_request = array(
-            'amount' => $this->getAmount(),
-            'card' => $this->getCardToken(),
-            'capture' => $this->getCapture(),
-            'currency' => $this->getCurrencyCode(),
-            'description' => $this->getChargeDescription(),
-        );
-
-        $secret_key = $this->getSecretKey();
+        $omiseCharge = new Charge();
 
         try {
-            $charge = OmiseCharge::create($charge_request, '', $secret_key);
+            $omiseCharge->create();
         } catch (Exception $e) {
             $this->context->smarty->assign('error_message', $e->getMessage());
         }
--- a/omise/controllers/front/payment.php
+++ b/omise/controllers/front/payment.php
@@ -4,6 +4,7 @@ if (! defined('_PS_VERSION_')) {
 }
 
 require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-php/lib/Omise.php';
+require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-plugin/Omise.php';
 require_once _PS_MODULE_DIR_ . '/omise/setting.php';
 
 class OmisePaymentModuleFrontController extends ModuleFrontController
@@ -15,42 +16,31 @@ class OmisePaymentModuleFrontController extends ModuleFrontController
     {
         parent::initContent();
 
-        $data = array(
+        $charge_request = array(
             'amount' => $this->getAmount(),
             'card' => $this->getCardToken(),
             'capture' => $this->getCapture(),
-            'currency' => $this->getCurrency(),
-            'description' => $this->getDescription(),
+            'currency' => $this->getCurrencyCode(),
+            'description' => $this->getChargeDescription(),
         );
 
         $secret_key = $this->getSecretKey();
 
         try {
-            $charge = OmiseCharge::create($data, '', $secret_key);
-            $payment_success = true;
+            $charge = OmiseCharge::create($charge_request, '', $secret_key);
         } catch (Exception $e) {
-            $payment_success = false;
-            $this->context->smarty->assign(
-                array(
-                    'error_message' => $e->getMessage(),
-                )
-            );
+            $this->context->smarty->assign('error_message', $e->getMessage());
         }
-
-        $this->context->smarty->assign(
-            array(
-                'payment_success' => $payment_success,
-            )
-        );
 
         $this->setTemplate('payment_result.tpl');
     }
 
     public function getAmount()
     {
-        $total = (float) $this->context->cart->getOrderTotal(true, Cart::BOTH);
+        $currency_code = $this->getCurrencyCode();
+        $order_total = (float) $this->context->cart->getOrderTotal(true, Cart::BOTH);
 
-        return 100 * $total;
+        return OmisePluginHelperCharge::amount($currency_code, $order_total);
     }
 
     public function getCardToken()
@@ -63,17 +53,17 @@ class OmisePaymentModuleFrontController extends ModuleFrontController
         return 'true';
     }
 
-    public function getCurrency()
+    public function getChargeDescription()
+    {
+        return 'Charge a card using a token from PrestaShop (' . _PS_VERSION_ . ')';
+    }
+
+    public function getCurrencyCode()
     {
         $currency_id = (int) $this->context->cart->id_currency;
         $currency_instance = Currency::getCurrencyInstance($currency_id);
 
         return $currency_instance->iso_code;
-    }
-
-    public function getDescription()
-    {
-        return 'PrestaShop';
     }
 
     public function getSecretKey()
--- a/omise/controllers/front/payment.php
+++ b/omise/controllers/front/payment.php
@@ -1,0 +1,85 @@
+<?php
+if (! defined('_PS_VERSION_')) {
+    exit();
+}
+
+require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-php/lib/Omise.php';
+require_once _PS_MODULE_DIR_ . '/omise/setting.php';
+
+class OmisePaymentModuleFrontController extends ModuleFrontController
+{
+    public $context;
+    public $display_column_left = false;
+
+    public function initContent()
+    {
+        parent::initContent();
+
+        $data = array(
+            'amount' => $this->getAmount(),
+            'card' => $this->getCardToken(),
+            'capture' => $this->getCapture(),
+            'currency' => $this->getCurrency(),
+            'description' => $this->getDescription(),
+        );
+
+        $secret_key = $this->getSecretKey();
+
+        try {
+            $charge = OmiseCharge::create($data, '', $secret_key);
+            $payment_success = true;
+        } catch (Exception $e) {
+            $payment_success = false;
+            $this->context->smarty->assign(
+                array(
+                    'error_message' => $e->getMessage(),
+                )
+            );
+        }
+
+        $this->context->smarty->assign(
+            array(
+                'payment_success' => $payment_success,
+            )
+        );
+
+        $this->setTemplate('payment_result.tpl');
+    }
+
+    public function getAmount()
+    {
+        $total = (float) $this->context->cart->getOrderTotal(true, Cart::BOTH);
+
+        return 100 * $total;
+    }
+
+    public function getCardToken()
+    {
+        return Tools::getValue('omise_card_token');
+    }
+
+    public function getCapture()
+    {
+        return 'true';
+    }
+
+    public function getCurrency()
+    {
+        $currency_id = (int) $this->context->cart->id_currency;
+        $currency_instance = Currency::getCurrencyInstance($currency_id);
+
+        return $currency_instance->iso_code;
+    }
+
+    public function getDescription()
+    {
+        return 'PrestaShop';
+    }
+
+    public function getSecretKey()
+    {
+        $setting = new Setting();
+
+        return $setting->getSecretKey();
+    }
+}
--- a/omise/controllers/front/payment.php
+++ b/omise/controllers/front/payment.php
@@ -3,10 +3,6 @@ if (! defined('_PS_VERSION_')) {
     exit();
 }
 
-require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-php/lib/Omise.php';
-require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-plugin/Omise.php';
-require_once _PS_MODULE_DIR_ . '/omise/setting.php';
-
 class OmisePaymentModuleFrontController extends ModuleFrontController
 {
     public $context;
--- a/omise/controllers/front/payment.php
+++ b/omise/controllers/front/payment.php
@@ -3,6 +3,8 @@ if (! defined('_PS_VERSION_')) {
     exit();
 }
 
+require_once _PS_MODULE_DIR_ . '/omise/classes/charge.php';
+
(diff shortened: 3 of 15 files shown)


omise/controllers/front/payment.php
```diff
@@ -0,0 +1,28 @@
+<?php
+if (! defined('_PS_VERSION_')) {
+    exit();
+}
+
+if (defined('_PS_MODULE_DIR_')) {
```
guzzilar, line 6: @nimid better to check if `_PS_MODULE_DIR_` is not defined, then, `exit();`? otherwise the code below won't work.

omise/controllers/front/payment.php
```diff
@@ -0,0 +1,28 @@
...
+    require_once _PS_MODULE_DIR_ . '/omise/classes/charge.php';
+}
+
```
guzzilar, line 9: @nimid for the delegate class design at `omise/classes/charge.php`
I think we can more separate their responsibilities by keep `omise/classes/charge.php` do only connect with Omise API without know how to gather data from a form.

It means, `omise/classes/charge.php` doesn't need to extends `ModuleFrontController` and let `OmisePaymentModuleFrontController` pass the values that this class gather from a checkout form by itself, since this class only the class that directly communicate with frontend.

omise/controllers/front/payment.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
guzzilar, line 1: @nimid if it's not break any PrestaShop rules, is it possible to have file name to `OmisePaymentModuleFrontController.php` to reflect its class name?

omise/classes/charge.php
```diff
@@ -0,0 +1,64 @@
+<?php
```
guzzilar, line 1: @nimid if it's not break any PrestaShop rules, is it possible to have file name to `Charge.php` to reflect its class name?

omise/classes/charge.php
```diff
@@ -0,0 +1,64 @@
...
+    exit();
+}
+
+if (defined('_PS_MODULE_DIR_')) {
+    require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-php/lib/Omise.php';
+    require_once _PS_MODULE_DIR_ . '/omise/libraries/omise-plugin/Omise.php';
+    require_once _PS_MODULE_DIR_ . '/omise/setting.php';
+}
+
+class Charge extends ModuleFrontController
```
guzzilar, line 12: @nimid maybe better to add some prefix name? to avoid duplicate class name (that some 3rd-party modules may have the same name, since it's common name).

omise/classes/charge.php
```diff
@@ -0,0 +1,64 @@
...
+
+        return OmisePluginHelperCharge::amount($currency_code, $order_total);
+    }
+
+    protected function getCardToken()
+    {
+        return Tools::getValue('omise_card_token');
+    }
+
+    protected function getCapture()
```
guzzilar: @nimid continue from https://github.com/omise/omise-prestashop/pull/19/files#r97012434
might be better to retrieve the `payment method` setting from `Setting` class instead?

omise/classes/charge.php
```diff
@@ -0,0 +1,64 @@
+<?php
+if (! defined('_PS_VERSION_')) {
+    exit();
+}
+
+if (defined('_PS_MODULE_DIR_')) {
```
guzzilar, line 6: @nimid same as https://github.com/omise/omise-prestashop/pull/19/files#r97011969

omise/controllers/front/payment.php
nimid: @guzzilar No, it is broken. The last part of the URL, `/payment`, is consistent with the file name.

The screenshot below shows the error result page after changed a file name from `payment.php` to `OmisePaymentModuleFrontController.php`.

404_error_-_omise-prestashop

Did you test and is it work on your environment?

omise/controllers/front/payment.php
nimid: @guzzilar The reason of this condition has been provided in the description of the commit, https://github.com/omise/omise-prestashop/commit/d0f5faea4f5a2438a7f307859c71d494bd376716.

omise/controllers/front/payment.php
nimid: @guzzilar To remove the extending `ModuleFrontController` at the class `Charge`, the command that used to get the PrestaShop context `$this->context` can be changed to `Context::getContext()`.

For example:

Before

```php
$this->context->cart->id_currency
```

After
```php
Context::getContext()->cart->id_currency
```

PrestaShop Reference: [How to access the Context?](http://doc.prestashop.com/display/PS15/Using+the+Context+Object#UsingtheContextObject-HowtoaccesstheContext?)

omise/classes/charge.php
nimid: @guzzilar The reason of this condition has been provided in the description of the commit, 0924c74f2c824d8cfa2e0e65d488daf1683a0072.

omise/classes/charge.php
nimid: @guzzilar The payment method will be developed along with the development of a feature, manual capture.

omise/controllers/front/payment.php
guzzilar: 👌 & nice explanation 👍

omise/controllers/front/payment.php
guzzilar: @nimid I mean, we can separate the responsibility of classes.
Since `OmisePaymentModuleFrontController` is a class that communicate between Front & Back. This class knows, what items are in a cart, which currency buyer's using, total amount of a cart.

And `Charge` knows Omise's format.

Then, code layer could be
```
       |                           |
       |                           |
Omise < > Charge                  < > OmisePaymentModuleFrontController
=======================================================================
       | • Formatting cart's info  | • Collect cart's informations
       |   to Omise format         | • Pass params to Charge
       | • OmiseCharge::create();  | • Charge::create(informations);
       |                           | 
```

nimid: @guzzilar At first, it has only one class, `OmisePaymentModuleFrontController`. [This class does all functional to create the charge](https://github.com/omise/omise-prestashop/blob/bc8b1d323ef63866574590989ee278a5d5170448/omise/controllers/front/payment.php). It has many dependencies and functionalities.

Along with the rise of the features that will be developed soon, the class `OmisePaymentModuleFrontController` will has more size, logic and complexity that adversely affect to the maintenance. For an example from another project, [function process_payment()](https://github.com/omise/omise-woocommerce/blob/master/omise-wc-gateway.php#L88).

So, a class `OmisePaymentModuleFrontController ` has been reduced the dependencies to be one dependency (`Charge`).

Including, PrestaShop has provided the internal functions to retrieve the information such as cart, currency and etc. These functions enable class `Charge` to retrieve all information that need to create Omise charge by itself. Which it does not need to pass from class `OmisePaymentModuleFrontController`.

The knowledge of how to charge has been packed in `Charge` class.
The controller class `OmisePaymentModuleFrontController` is control the flow.

How about a class `Charge` has been renamed to `OmiseChargeModuleFrontController`?

guzzilar: @nimid I agree about separate classes idea.
Btw from this design, in fact, the dependencies weren't removed. It just moves from `OmisePaymentModuleFrontController` to `Charge`.

What I meant is, we can do more decoupling by, let `Controller` take cares only `input`. And then, `Charge` take cares "How the class gonna communicate to Omise API and return back to controller" (i.e. amount formatting.)

For example
Controller just handle "input" and pass those input to `Charge`
```php
// OmisePaymentModuleFrontController.php

public function initContent()
{
    parent::initContent();

    try {
        $charge = new Charge;
        $charge->create(array(
            'amount'      => $this->context->cart->orderTotal(),
            'card'        => Tools::getValue('omise_card_token'),
            'currency'    => $this->context->currency->iso_code,
            'description' => "...",
        ));
    } catch (Exception $e) {
        $this->context->smarty->assign('error_message', $e->getMessage());
    }

    $this->setTemplate('payment_result.tpl');
}
```

And Charge just care about formatting stuff, and how to communicate to Omise API
```php
// Charge.php

public function create($params)
{
    $setting = new Setting();

    switch ($setting->payment_action) {
        case 'manual_capture':
            return $this->authorize($params, $setting->getSecretKey());
            break;

        case 'auto_capture':
            return $this->capture($params, $setting->getSecretKey());
            break;
    }

    return // fail, unknow command.
}

public function authorize($params, $secret_key)
{
    $params['amount']  = OmisePluginHelperCharge::amount($params['currency'], $params['amount']);
    $params['capture'] = false;

    return OmiseCharge::create($params, '', $secret_key);
}

public function capture($params, $secret_key)
{
    $params['amount']  = OmisePluginHelperCharge::amount($params['currency'], $params['amount']);
    $params['capture'] = true;

    return OmiseCharge::create($params, '', $secret_key);
}
```

From this design, let says, if we have "payment_action" == "internet banking" (just assumed we're gonna store "internet banking" option into `payment_action`

We just have another method called `internetBanking() { ... }` and then, this method will do stuff to communicate with Omise API.

guzzilar: fyi, I found that we can get `iso_code` from `$this->context->currency` in a controller class.

nimid: @guzzilar Now, the discussion is should `OmisePaymentModuleFrontController` class send the parameter to `Charge.create()`?

From the above example:

- The coupling between `OmisePaymentModuleFrontController ` and `Charge` is increased with an array parameter that contained 4 parameters inside.

- The dependency of `OmisePaymentModuleFrontController ` is increased more 3 (`Tools`, `Cart` and `Currency`). Which it return to the original problem that required refactoring.

- The parameter that used to send to Omise API has been separated to 2 places. Some in `OmisePaymentModuleFrontController` and some in `Charge`.

- If the parameter that need to send to Omise API has been changed, it will impact through to the controller class. It is not be scoped in one class or one place.

guzzilar: @nimid how about try to think in MVC pattern?

Assumed, 
- `OmisePaymentModuleFrontController` = Controller
- `Charge` = Model
- `OmiseCharge` = DB

Should model care about how to retrieve input by itself (by retrieving `$_POST['card']` inside a model) or better to pass through controller?

guzzilar: Btw, dependency of `OmisePaymentModuleFrontController` doesn't increase.
It's what "ControllerCore" provided for us (from what we extended `ModuleFrontController`).

Instead, if we call `Currency::xxx()`, `Cart::xxx()` inside another class that doesn't extend `ModuleFrontController` directly, that's increase dependency.

nimid: @guzzilar `Charge` is controller that has been delegated for `OmisePaymentModuleFrontController`.

nimid: If it has calling, both static call and extending are the dependency.

The unit testing is help to emphasize the dependency.

guzzilar: @nimid if charge is a controller, then it's no reason to have `flow controller` as `OmisePaymentModuleFrontController` isn't it?

If you care about "Along with the rise of the features that will be developed soon, the class OmisePaymentModuleFrontController will has more size, logic and complexity that adversely affect to the maintenance. For an example from another project, function process_payment()."

Then, just have more controller with multiple endpoints.

i.e.
- controllers/charge.php
- controllers/refund.php
- controllers/threedsecure.php
- controllers/internetbanking.php

I don't see the benefit of have 2 controllers (delegated class & flow controller class).

Say, if you have to develop more feature in the future, your `charge.php` will "has more size, logic and complexity that adversely affect to the maintenance" anyway, since that `flow controller class` do nothing, just bypass action to another delegated class.

---

Btw, to reduce confusing, if `charge.php` is a controller, could you move it to `controllers` folder?

---

> "If it has calling, both static call and extending are the dependency."

Yup, we have no way to retrieve "orderTotal()" if we are not using `Cart` dependency.
But the question is, if `ModuleFrontController` already provided by itself, why don't we just use `$this->context->cart` from itself?

👆 btw, this one may be a totally different topic, I have this idea because at first I confused and thought that `classes/charge.php` is a model-like class.

guzzilar: @nimid Im thinking in a way that `classes/charge.php` is a controller,

Now understand that why you want to have "get*()" inside `Charge.php` itself.

But then, now I don't see the benefit to have delegated class.
It just did move "logic & stuff" from main class to delegated class, but at the end, what you've concerned will happen at that delegated class anyway.

nimid: @guzzilar The complexity has been reduced.

From https://github.com/omise/omise-prestashop/pull/19#discussion_r97259710

> The knowledge of how to charge has been packed in `Charge` class.

> The controller class `OmisePaymentModuleFrontController` is control the flow.

The change below is out of scope of this pull request but it can make more understanding.

https://github.com/omise/omise-prestashop/blob/ba59c62cd78411fb07655e3dc5932f9daef50e51/omise/controllers/front/payment.php

guzzilar: ok, last question about this design (and it'd not relate with what we've talked above).

Because this design moved code into `Charge` controller, so, it hard to reuse the class (i.e. if we implement "Refund" service).

I'd suggest to 
- make `OmisePaymentModuleFrontController` as an abstract method that extend `ModuleFrontController`. 
- Then, `Charge` will extends `OmisePaymentModuleFrontController`.
- Then, we implement `postProcess` method in `Charge.php` (and update endpoint of submit payment to this class instead)

- Then, provide some basic functions that we could reuse in every controller classes to `OmisePaymentModuleFrontController`
i.e. 
```php
abstract class OmisePaymentModuleFrontController extends ModuleFrontController
{
    protected $setting;

    public function __construct()
    {
        $this->setting = new Setting();
    }

    public function redirect()
    {
        // Redirect to the result/ 3-D Secure/ Internet Banking page.
    }

    public function handleFailure()
    {
        // - Update order status to "failed",
        // - Redirect to fail page
        // - Etc.
    }
...
```

- Now you can use `$this->context` class inside `Charge`, and every controller class in the future can also extend `OmisePaymentModuleFrontController` to use `$this->context`.

- Then, in `Charge` Controller, just provide only method that can't reuse. (i.e. `$this->getCardToken()`)

- So, in `Charge` controller could be

```php
class Charge extends OmisePaymentModuleFrontController
{
    public function postProcess()
    {
        $charge_request = array(
            'amount'      => $this->getAmount(),
            'card'        => $this->getCardToken(),
            'capture'     => $this->getCapture(),
            'currency'    => $this->context->currency->iso_code ,
            'description' => $this->getChargeDescription(),
        );

        try {
            OmiseCharge::create($charge_request, '', $this->setting->getSecretKey());

            $this->setRedirectAfter('success-page');
        } catch(Exception $e) {
            $this->handleFailure();
        }
    }

...
```

☝️ I know this idea will not solve "decoupling" class issue as you want much.

But since PrestaShop has been provided `postProcess` for us, each class (i.e. `Charge.php`, `Refund.php`) just focus on implement their own `postProcess()` method, and lets handler methods live in `OmisePaymentModuleFrontController`

It'd make `OmisePaymentModuleFrontController` can be reusable and reduce duplicated code in case we have another controller for another service (i.e. refund).

guzzilar: The reasons behind this design are, 
- To make code reusable on `OmisePaymentModuleFrontController`
- To make `Charge` class conform with `single responsibility` principle by, 1 class will take care only 1 thing (which is, `postProcess` function.)
- Prevent duplicating code in the future if we have more controller.
- And because I found that PrestaShop provided `postProcess()` & `redirect()` methods, that made our code more flexible to design.

guzzilar:
@nimid 
I checked at PR #21.
I think that design is a bit complex,

- OmiseBasePaymentModuleFrontController shouldn't contain logic in `postProcess`
    If this base class contains logic to create charge itself, then it shouldn't named as `base` cuz it can't reuse for other purpose except for create charge

- To make OmiseBasePaymentModuleFrontController more flexible.
    I think OmiseBasePaymentModuleFrontController shouldn't handle `postProcess` itself, instead, `controllers/front/payment.php` or `controllers/front/threedomainsecurepayment.php` should decide by themselves which action they're going to perform.

- Continue from above, then we can move `try { } catch() { }` into `classes/charge.php` instead.
    Since `classes/charge.php` performs `create` by itself, I think it's better to let that class handle `catch exception`.

- Slightly, about `OmiseThreeDomainSecure` name.
    Do you have any reason behind that you chose `OmiseThreeDomainSecure` instead of `OmiseThreeDSecure`?
    Asking because usually in Omise plugin, we keep exposing term of `ThreeDSecure` instead of `ThreeDomainSecure`.
    If there is no special reason behind, I prefer to use the same name as other plugins instead.

nimid:
@guzzilar

- Can you provide other use cases at the front-end side (by the payer) except create charge?


How about an alternative design below?

- A function `Charge.createInternetBanking()` is for example only.

```php
class OmisePaymentModuleFrontController extends OmiseBasePaymentModuleFrontController
{
    public function postProcess()
    {
        try {
            $this->charge = $this->omise_charge->create(Tools::getValue('omise_card_token'));
        } catch (Exception $e) {
            $this->error_message = $e->getMessage();
            return;
        }

        if ($this->charge->isFailed()) {
            $this->error_message = $this->charge->getErrorMessage();
            return;
        }

        $this->payment_order->save();

        $this->setRedirectAfter('index.php?controller=order-confirmation' .
            '&id_cart=' . $this->context->cart->id .
            '&id_module=' . $this->module->id .
            '&id_order=' . $this->module->currentOrder .
            '&key=' . $this->context->customer->secure_key
        );
    }
}
```

```php
class OmiseThreeDomainSecurePaymentModuleFrontController extends OmiseBasePaymentModuleFrontController
{
    public function postProcess()
    {
        $this->payment_order->saveAsProcessing();

        try {
            $this->charge = $this->omise_charge->createThreeDomainSecure(Tools::getValue('omise_card_token'));
        } catch (Exception $e) {
            $this->error_message = $e->getMessage();
            return;
        }

        if ($this->charge->isFailed()) {
            $this->error_message = $this->charge->getErrorMessage();
            return;
        }

        $this->setRedirectAfter($this->charge->getAuthorizeUri());
    }
}
```

```php
abstract class OmiseBasePaymentModuleFrontController extends ModuleFrontController
{
    protected $charge;
    public $display_column_left = false;
    protected $error_message;
    protected $omise_charge;
    protected $payment_order;
    protected $setting;

    public function __construct()
    {
        parent::__construct();

        $this->omise_charge = new Charge();
        $this->payment_order = new PaymentOrder();
        $this->setting = new Setting();
    }

    public function initContent()
    {
        parent::initContent();

        $this->context->smarty->assign('error_message', $this->error_message);
        $this->setTemplate('payment-error.tpl');
    }

    public function redirect()
    {
        Tools::redirect($this->redirect_after);
    }
}
```

```php
class Charge
{
    ...

    public function create($card_token)
    {
        $charge_request = array(
            'amount' => $this->getAmount(),
            'card' => $card_token,
            'capture' => $this->getCapture(),
            'currency' => $this->getCurrencyCode(),
            'description' => $this->getChargeDescription(),
        );

        $this->charge_response = OmiseCharge::create($charge_request, '', $this->getSecretKey());

        return $this;
    }

    public function createInternetBanking()
    {
        $charge_request = array(
            'amount' => $this->getAmount(),
            'capture' => $this->getCapture(),
            'currency' => $this->getCurrencyCode(),
            'description' => $this->getChargeDescription(),
            'offsite' => 'internet_banking_scb',
            'return_uri' => $this->getReturnUri(),
        );

        $this->charge_response = OmiseCharge::create($charge_request, '', $this->getSecretKey());

        return $this;
    }

    public function createThreeDomainSecure($card_token)
    {
        $charge_request = array(
            'amount' => $this->getAmount(),
            'card' => $card_token,
            'capture' => $this->getCapture(),
            'currency' => $this->getCurrencyCode(),
            'description' => $this->getChargeDescription(),
            'return_uri' => $this->getReturnUri(),
        );

        $this->charge_response = OmiseCharge::create($charge_request, '', $this->getSecretKey());

        return $this;
    }

...

}
```

> Do you have any reason behind that you chose `OmiseThreeDomainSecure` instead of `OmiseThreeDSecure`?

I found the full word for explanation can have more clearly understandable. If a word `Domain` is long name, the abbreviation is preferred but for more 5 characters (`omain`), it can help in part of understandable.

In the camel case coding convention, when `D` beside `Secure`, it may make the confusion.

composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require-dev": {
+        "mockery/mockery": "dev-master"
+    }
+}
```
guzzilar: @nimid note, you can have 
```json
"autoload": {
    "classmap": ["omise/"],
    "exclude-from-classmap": ["/omise/libraries/"]
}
```
That you can get rid of maintaining `tests/autoload.php`, `function autoload($class) { ... }` function.

omise/classes/charge.php
guzzilar, line 1: @nimid can you provide docblock for 
- create()
- getAmount()
- getCardToken()
- getCapture()
- getChargeDescription()
- getCurrencyCode()
- getSecretKey()

composer.json
nimid: @guzzilar Done. 2db48203a39f7f38d57398e90b950d62c48c7dfe

omise/classes/charge.php
nimid: @guzzilar Are there other parts that need to be changed?

To merge the pull request, Is it waiting for only DocBlocks update (https://github.com/omise/omise-prestashop/pull/19#discussion_r102140277)?

guzzilar: @nimid Checking. But thought nowhere to comment more. Going to finalise soon-ish

tests/unit/classes/ChargeTest.php
```diff
@@ -0,0 +1,90 @@
...
+            ->shouldReceive('create')
+            ->with($this->createChargeRequest(), '', $this->secret_key)
+            ->once();
+
+        $this->charge->create();
+    }
+
+    private function createChargeRequest()
+    {
+        $charge_request = array(
```
guzzilar, line 64: @nimid you can just return array here.

tests/unit/classes/ChargeTest.php
```diff
@@ -0,0 +1,90 @@
...
+            'card' => 'cardToken',
+            'capture' => 'true',
+            'currency' => 'THB',
+            'description' => 'Charge a card using a token from PrestaShop (' . _PS_VERSION_ . ')',
+        );
+
+        return $charge_request;
+    }
+
+    public function tearDown() {
```
guzzilar, line 75: @nimid pls have 1 break line for `{`

guzzilar:
@nimid 
_Note, I reviewed without checking another PR, so, some of my comments may already been solved in somewhere else in some PRs._
_(And also, we may have already talked about it in some part of my comments. 
Feel free to point it out if comments below were mentioned/duplicated in some place even in this PR)_

Here we go.

- Docblock(s) are missing.
    https://github.com/omise/omise-prestashop/pull/19/files#r102140277

- If `classes/charge.php` is a controller, I think it better to keep it in `controllers` folder.

- `classes/charge.php` shall we rename it or make it more unique name? (i.e. OmiseCharge.php) to avoid duplicate class name (that some 3rd-party modules may have the same name, since it's pretty common name).
    https://github.com/omise/omise-prestashop/pull/19/files#r97012652

- Personally, I try avoid using `getter`, `setter` functions without specific case and just use it when I really need to add some complex logic in it.
    i.e. [classes/charge.php, getAmount()](https://github.com/omise/omise-prestashop/pull/19/files#diff-d292a4e7ca9f76a040043bc658e976c4R27) looks fine for me but [classes/charge.php, getCapture()](https://github.com/omise/omise-prestashop/pull/19/files#diff-d292a4e7ca9f76a040043bc658e976c4R40) looks unnecessary and it makes function name looks weird. 
    (think about interface. `Charge->getCapture()`. My first impression when I saw it. It's totally nothing deal with `Payment Action: Authorize only or Authorize and Capture`. Because `getCapture` is meaningless.

- In `classes/charge.php` if you separate between `create()` and `buildRequest()`. I think it's gonna be easier for test.
    i.e.
    ```php
    class Charge extends ModuleFrontController
    {
        /**
         * @return OmiseCharge
         */
        public function create()
        {
            return OmiseCharge::create($this->buildRequest(), '', $this->getSecretKey());
        }

        /**
         * @return array
         */
        protected function buildRequest()
        {
            return array(
                'amount'      => $this->getAmount(),
                'card'        => $this->getCardToken(),
                'capture'     => $this->getCapture(),
                'currency'    => $this->getCurrencyCode(),
                'description' => $this->getChargeDescription(),
            );
        }
    }
    ```

- Vertical alignment would make your code much easier to read.

nimid:
@guzzilar https://github.com/omise/omise-prestashop/pull/19#issuecomment-281457688

> Docblock(s) are missing.

Which [DocBlocks tag](https://phpdoc.org/docs/latest/internals/tags.html) is required?

<br />

> If classes/charge.php is a controller, I think it better to keep it in controllers folder.

It has been [changed](https://github.com/omise/omise-prestashop/pull/21/files#diff-d292a4e7ca9f76a040043bc658e976c4R12).

<br />

> classes/charge.php shall we rename it or make it more unique name?

`OmiseCharge` has [already defined](https://github.com/omise/omise-prestashop/blob/feature/make-charge-with-omise/omise/libraries/omise-php/lib/omise/OmiseCharge.php#L6).

How about `OmiseChargeClass`?

<br />

> but classes/charge.php, getCapture() looks unnecessary and it makes function name looks weird.

Done. 672b59305ef127367c4c5fe7c4673e3ba8a7c56d

<br />

> I think it's gonna be easier for test.

Can you show the test, compare and clarify how it easier?

<br />

> Vertical alignment would make your code much easier to read.

It is not my code. It is not your code. It is not our code.

**It belongs to Omise.**

It is not just change the ownership words. It is the attitude in working. It is a way how to behave with it. It is the attitude when communicate with co-worker and people.

I found you have shared a [code review guideline](https://github.com/thoughtbot/guides/tree/master/code-review). In the guideline, it clarified *Avoid selective ownership of code. ("mine", "not mine", "yours")*.

Is vertical alignment personal preference?

If it is not, can you show the publicly accepted standard for reference that this project should be consistent?

tests/unit/classes/ChargeTest.php
nimid: @guzzilar Can you clarify?

tests/unit/classes/ChargeTest.php
nimid: @guzzilar It is done. https://github.com/omise/omise-prestashop/blob/feature/3ds/tests/unit/classes/ChargeTest.php#L173

nimid:
The source code in this pull request has been changed and combined to merge together with the pull request #21.

PR #59, opened 2018-06-05
Removes mention of specific internet banking charges

jonrandy:
#### 1. Objective

Replace labels referring to internet banking charges with a general warning that they may be charged a small fee by their bank for these payments

#### 2. Description of change

Modified the relevant template, and css file

#### 3. Quality assurance

Set up a sample store, installed plugin, switch on internet banking payments, viewed checkout page

#### 4. Impact of the change

The internet banking section now has a general warning about bank fees instead of specific fee details.
image

#### 5. Priority of change

Normal

#### 6. Additional notes

🎶

### DIFF
--- a/omise/css/omise_internet_banking.css
+++ b/omise/css/omise_internet_banking.css
@@ -27,7 +27,7 @@
   width: 30px;
   height: 30px;
 }
-.omise-internet-banking .omise-bank-text-wrapper { display: inline-block; line-height:38px; }
+.omise-internet-banking .title { line-height:38px; }
 .omise-internet-banking .scb { background: #4e2e7f none repeat scroll 0 0; }
 .omise-internet-banking .ktb { background: #1ba5e1 none repeat scroll 0 0; }
 .omise-internet-banking .bay { background: #fec43b none repeat scroll 0 0; }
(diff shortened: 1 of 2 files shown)


omise/css/omise_internet_banking.css
```diff
@@ -27,11 +27,7 @@
   width: 30px;
   height: 30px;
 }
-.omise-internet-banking .omise-bank-text-wrapper { display: inline-block; }
-.omise-internet-banking .secondary-text {
-  font-size: 80%;
-  color: #aaa;
-}
+.omise-internet-banking .omise-bank-text-wrapper { display: inline-block; line-height:38px; }
```
guzzilar: @jonrandy would it better to set `line-height` inside `span.title`? 
something like 👇 
```css
.omise-bank-text-wrapper .title { line-height:38px; }
```

Just thought that `display: inline-block` is unnecessary here (maybe)

jonrandy: I reworked the html a bit now, there was some `div`-itis going on there. Less HTML and CSS now, same result

jonrandy:
@guzzilar Better?

guzzilar (approved):
👍

PR #61, opened 2018-06-26
Composer omise php

jonrandy:
#### 1. Objective

Make it easier to maintain the PrestaShop plugin by pulling in `omise-php` using composer, instead of just having a 'fixed' copy of a particular version of that library stored in the plugin's repo.

#### 2. Description of change

Completely removed the `omise-php` library from the repo, added a `composer.json` file, and modified code so that the path to the now 'composer managed' files is correct.

(I also fixed the tests as they were not runnable with recent versions of PHPUnit)

#### 3. Quality assurance

Checked to make sure the plugin still works, which it does

#### 4. Impact of the change

No impact for users as nothing actually changes. Impact is ease of maintenance for devs

#### 5. Priority of change

Normal

#### 6. Additional notes

Calm down England. It was only Panama, it's not like we won the final!


guzzilar:
@jonrandy how about the installation process?
Right now we guide user to download the repo as a zip file n upload directly into their store.

jonrandy:
Uploading a zip is AFAIK the recommended process in PrestaShop. I could write a small shell script to help devs easily produce the required zip file?

jonrandy:
Unless of course we add the plugin to the PrestaShop 'store'

jonrandy:
This PR doesn't affect the install process as the ZIP files for the user to download that we add on the releases are manually created by us. They aren't the same as the source code zip file

jonrandy:
So, is this one ok?

jacstn:
I think we need more info in _How to test it.._

jonrandy:
That would depend on how you wanted to set up your own dev environment. Personally I have a dockerized solution with the relevant plugin folder inside my container pointing at the `omise` folder in my local prestashop dev folder. Then I can run `composer install` to bring in the libs, then test to see if the plugin works as expected.

Here is my `docker_compose.yml` if you want to set it up that way.

```yaml
version: '2'

services:
  mariadb:
    image: 'bitnami/mariadb:latest'
    environment:
      - MARIADB_USER=bn_prestashop
      - MARIADB_DATABASE=bitnami_prestashop
      - ALLOW_EMPTY_PASSWORD=yes
    volumes:
      - 'mariadb_data:/bitnami'
  prestashop:
    image: 'bitnami/prestashop:latest'
    environment:
      - MARIADB_HOST=mariadb
      - MARIADB_PORT_NUMBER=3306
      - PRESTASHOP_HOST=localhost
      - PRESTASHOP_DATABASE_USER=bn_prestashop
      - PRESTASHOP_DATABASE_NAME=bitnami_prestashop
      - ALLOW_EMPTY_PASSWORD=yes
    labels:
      kompose.service.type: nodeport
    ports:
      - '80:80'
      - '443:443'
    volumes:
      - './prestashop_data:/bitnami'
      - '../omise-prestashop/omise:/bitnami/prestashop/modules/omise'
    depends_on:
      - mariadb

volumes:
  mariadb_data:
    driver: local
  prestashop_data:
    driver: local

```

.gitignore
```diff
@@ -1,4 +1,5 @@
 .buildpath
 .project
-.settings/
+/omise/vendor
```
guzzilar, line 3: @jonrandy is this should be `vendor/omise`?

guzzilar: ah sry ic, we have `omise` folder at the root dir. 😂

guzzilar (changes requested):
The changes looks fine.
Just need to update README.md at the Installation Instruction to make it conform with the change.

Also, need to make sure that we have a `built` version after we release for the next version v1.6 (it's ok for now since our document pointed to the released-file. but after we release this one, gotta need to make sure that end-user can install it without running `composer install`)

jonrandy:
@guzzilar The installation instructions are the same. We have always built the zip,
and that's what they install. This is merely a change to make our
(developers) lives easier

guzzilar (approved):
@jonrandy according to
>  The installation instructions are the same. We have always built the zip,

So I guess we need to make a conclusion on where to store that zip file, and update all documents to point to that link instead?

PR #62, opened 2018-07-13
Dual version compatibility

jonrandy:
#### 1. Objective

This PR reinstates compatibility with Prestashop 1.6, which appears to be far more popular than 1.7. It does not however, break 1.7 compatility - we now have a plugin that works on both versions. Merchants who use 1.6 will be able to enjoy all future (and past - there have been 2 releases without 1.6 compatibility) enhancements to the plugin.

#### 2. Description of change

Essentially I compared the 2 versions, and wrote code that cater to either version in areas where version specific code is required. The changes were surprisingly not that widespread. Luckily, Prestashop had documented the changes from 1.6 -> 1.7 quite well [here](http://build.prestashop.com/news/module-development-changes-in-17/)

#### 3. Quality assurance

I tested using Bitnami's docker images for PrestaShop from [here](https://github.com/bitnami/bitnami-docker-prestashop).

Tested with Prestshop versions `1.7.4` and `1.6.1.9`

**Details:**

To test the plugin, you should set up a Prestashop instance for each version (1.6, 1.7), install the plugin, and check all plugin features are working. I recommend using the Docker method above - super easy

N.B. - there seems to be an issue with Prestashop 1.7 whereby the installer appears to never end. However, this does not appear to be our fault as I've tried installing a number of other plugins and this occurs with all of them. If you refresh the page after the installer appears to have hung, you will find the plugin has actually installed successfully. I've googled this issue, but as yet have found no resolution.

#### 4. Impact of the change

Prestashop 1.6 users can now enjoy all the latest and greatest updates from us 😄 

#### 5. Priority of change

Normal

#### 6. Additional notes

Go Croatia! 🇭🇷 ⚽️ 🏆


jonrandy:
After we merge this, I think we can go for a new release `v1.6`

guzzilar:
@jonrandy is it possible to check version at the class-loader place instead of adding that `IS_VERSION_17` condition everywhere? 

Could be a place like 👇 ?
https://github.com/omise/omise-prestashop/blob/dual-version/omise/controllers/front/base.php#L6

And then maybe we can have to 2 classes to handle 2 version concerns separately.
```
- /omise/classes/1.6/omise_charge_class.php
- /omise/classes/1.7/omise_charge_class.php
```

---
More crazy idea is to have a new loader class to taking care of which files that needed to be loaded for a specific version.

guzzilar:
@jonrandy is it possible to check version at the class-loader place instead of adding that `IS_VERSION_17` condition everywhere? 

Could be a place like 👇 ?
https://github.com/omise/omise-prestashop/blob/dual-version/omise/controllers/front/base.php#L6

And then maybe we can have to 2 classes to handle 2 version concerns separately.
```
- /omise/classes/1.6/omise_charge_class.php
- /omise/classes/1.7/omise_charge_class.php
```

---
More crazy idea is to have a new loader class to taking care of which files that needed to be loaded for a specific version.

jonrandy:
Refactored to remove all those `if`s

omise/omise.php
```diff
@@ -290,6 +331,32 @@ public function install()
...
+    /**
+     * Register/Unregister all hooks for the module
+     *
+     * @return bool
+     */
+    protected function registerHooks($isRegister = true)
+    {
+        $res = true;
+        foreach (explode(',', PRESTASHOP_PAYMENTMODULE_HOOKS) as $hook) {
+            $res = $res && $this->{$isRegister ? "registerHook" : "unregisterHook"}($hook);
```
guzzilar: @jonrandy why don't we separate this method to `registerHooks` and `unregisterHooks` 🤔

guzzilar: I mean, it's strange that `registerHooks` method needed to decide whether executing `registerHook` or `unregisterHook` depending on `isRegister` is true or false.

jonrandy: Both methods would be all but identical - just the name of the method called would change. Pointless. DRY

jonrandy: I could change the name of the method to "changeHookRegistration" 😛

guzzilar: @jonrandy isn't that a clearer n meaningful name would make code more sense n easy to read? 🤔 

Compare between
```php
return parent::uninstall() && $this->registerHooks(false);
```

and

```php
return parent::uninstall() && $this->unregisterHooks();
```

jonrandy: Maybe, but it adds an unnecessary method, and repeated code. Or a tiny convenience method:

```php
function unregisterHooks() {
   $this->registerHooks(false);
}
```

In my mind, a total waste of time, and less efficient - 2 function calls instead of 1

jonrandy: Another way would be to pass `$this->unregisterHook` or `$this->registerHook` into a function `modifyHooks` which would loop through the hooks and run the passed function on them. Still avoiding the repeated code, or double function call. I'm not sure how easy passing functions around is with PHP though :)

guzzilar: @jonrandy  For the above, how about
```php
/**
 * Register/Unregister all hooks for the module
 *
 * @param  string $mode  Either 'registerHook' or 'unregisterHook'
 *
 * @return bool
 */
protected function toggleModuleHookRegistration($mode) {
    foreach (explode(',', PRESTASHOP_PAYMENTMODULE_HOOKS) as $hook) {
        if (! $this->{$mode}($hook)) return false;
    }

    return true;
}

/**
 * Register all hooks for the module
 *
 * @return bool
 */
protected function registerHooks()
{
    return $this->toggleModuleHookRegistration("registerHook");
}

/**
 * Unregister all hooks for the module
 *
 * @return bool
 */
protected function unregisterHooks()
{
    return $this->toggleModuleHookRegistration("unregisterHook");
}
```

This introduces on both readability and DRY, and code is a bit more explicit.

We might able to make it even more better by decoupling `PRESTASHOP_PAYMENTMODULE_HOOKS` out from the `toggleModuleHookRegistration` method and passing the module list into `registerHooks($modules)` and `unregisterHooks($modules)` instead (not necessary but it might help improving on the testability part as well, since we decoupling that constant out from the method.)

_note, just idea. I didn't test if this way work_

guzzilar: Understand that few more lines or methods aren't really necessary as the program is still function properly.
But if there is a way to improve its readability and it won't hurt anything (much), that would really helpful for any developers that are going to maintain or improve this code after us.

```php
return parent::uninstall() && $this->registerHooks(false);
```

This really doesn't sound right, as you can `unregister` hooks by passing `false` at the first argument of the method that named `registerHook`.

Don't think this thing is pointless thou, let's meet half-way 😉
(well, change method name as you suggested would work too, `changeHookRegistration` or something, but that way the code inside still a bit unclean as you have to check `$res = $res && $this->{$isRegister ? "registerHook" : "unregisterHook"}($hook);`)

omise/classes/omise_charge_class.php
```diff
@@ -9,6 +9,7 @@
     require_once _PS_MODULE_DIR_ . 'omise/setting.php';
 }
 
+
```
guzzilar: @jonrandy there is 1 accidentally breakline here

omise/config.xml
```diff
@@ -1,13 +1,13 @@
...
-	<confirmUninstall><![CDATA[Are you sure you want to uninstall Omise module?]]></confirmUninstall>
-    <is_configurable>1</is_configurable>
-    <need_instance>0</need_instance>
+	<name>omise</name>
+	<displayName><![CDATA[Omise]]></displayName>
+	<version><![CDATA[1.6]]></version>
+	<description><![CDATA[]]></description>
+	<author><![CDATA[Omise]]></author>
+	<tab><![CDATA[payments_gateways]]></tab>
+	<confirmUninstall><![CDATA[Are you sure you want to uninstall the Omise module?]]></confirmUninstall>
```
guzzilar, line 9: @jonrandy should we convert this line to `space` instead?

omise/omise.php
```diff
@@ -233,17 +252,18 @@ protected function getWebhooksEndpoint()
 
     public function hookDisplayOrderConfirmation($params)
     {
+
```
guzzilar: @jonrandy there is 1 accidentally breakline here

omise/omise.php
```diff
@@ -290,6 +331,32 @@ public function install()
...
 
+    /**
+     * A PrestaShop version specific version of 'display'
+     *
+     * @return see parent 'display' method
+     */
+    protected function versionSpecificDisplay($file, $template)
+    {
+        $versionedTemplate = PRESTASHOP_VERSION_VIEW_PATH . $template;
+        return $this->display($file, $versionedTemplate);
```
guzzilar: @jonrandy maybe `$versionedTemplate` variable is not really necessary here? How about just simply return
```php
return $this->display($file, PRESTASHOP_VERSION_VIEW_PATH . $template);
```

guzzilar:
@jonrandy I'm sorry for Croatia but saw you also tweaked a bit on the inapplicable template, can you also add a screenshot at the PR's description?

Thanks.. and sorry for Croatia again.. 😢

omise/omise.php
jonrandy: What you suggested is essentially identical to what I suggested (i.e. passing a function into a method that iterates over the hooknames and applies that function on them) except you are passing a method name instead of a function. The ultimate flexibility would obviously be gained by passing in a function, although I've never attempted that in PHP and some quick reading suggests it might be more trouble than it's worth, so passing in a string with a named method is the next best thing. I planned to do this tomorrow anyway (rename function and pass in a method name), but I won't add the additonal `registerHooks` and `unregisterHooks` methods as they're entirely unnecessary. I'll probably experiment with the passing a function method first though as I'm curious how this is done in PHP, and I like to program this way so I may use it more in the future :stuck_out_tongue: 

As for passing in the hook names, we could, but it is essentially pointless as the hooks are 'fixed' for the plugin - one set for each PrestaShop version. They don't change, and the only time anything is done with them is during the install/uninstall, where all hooks are registered/unregistered. I would change the method if a situation occurred that needed that change. As it stands, no such situation exists so there is no reason to change it. Coding for every possible eventuality is an exercise in futility. The method is as flexible as it needs to be for now.

As for testability - I've never written a test for any 'real' code in my entire life, I've only written tests as part of 'learning how to do automated testing' workshops. As far as I'm concerned, it is a simple method, and I _know_ it works. If I change it, and it breaks - I will see that when I am testing the plugin manually, and I will fix it before putting it live. Writing an automated test is unnecessary. That's how I see it anyway

omise/config.xml
jonrandy: Can do. I didn't consciously change it. Must have been my editor

omise/omise.php
jonrandy: Really??

omise/omise.php
jonrandy: OK

jonrandy:
@guzzilar There is no visible difference on either PrestaShop version that results from this tweak to the template

omise/omise.php
guzzilar: @jonrandy go for it 👍
Let me know if you want me to review the code again after finish experimenting that passing function thing :D

omise/omise.php
guzzilar: Oh, so it's an intentionally breakline?

guzzilar:
@jonrandy so this line is just for dual version compatibility? https://github.com/omise/omise-prestashop/pull/62/files#diff-374fd4fbc70b1795ff22dbb8ea4a8526R7

jonrandy:
Yes - the template between the older version of the plugin and the newer one only had one header tag removed. DIdn't seem worth splitting into 2 templates so I just made the template optionally accept a title

jonrandy:
Yes - the template between the older version of the plugin and the newer one only had one header tag removed. DIdn't seem worth splitting into 2 templates so I just made the template optionally accept a title

omise/omise.php
```diff
@@ -290,6 +335,30 @@ public function install()
...
+    /**
+     * Register/Unregister all hooks for the module
+     *
+     * @return bool
+     */
+    protected function applyToHooks($callable)
+    {
+        $res = true;
+        foreach (explode(',', PRESTASHOP_PAYMENTMODULE_HOOKS) as $hook) {
+            if (!$res &= call_user_func($callable, $hook)) break;
```
guzzilar, line 352: @jonrandy just wonder, can we just simply return `false` if the result of `call_user_func($callable, $hook))` is not `true`? (also, seems `if (!$res &= call_user_func($callable, $hook)) break;` converts the boolean result to `0`, `1`.)

omise/config.xml
jonrandy: I changed it, and PrestaShop itself changed it back. I'll leave it

guzzilar (approved):
@jonrandy  👍 Thanks for making this one 
I just left one more comment at  https://github.com/omise/omise-prestashop/pull/62/files#r202912179 

But the rest is 👍

omise/omise.php
jonrandy: That's exactly what it does - I just prefer returns to be at the end of a function if possible, not in the middle. I've switched back to a boolean although not necessary as 0 and 1 are truthy and falsey

guzzilar: @jonrandy then I guess `@return bool` is a bit misleading as the return type is not exactly boolean. Can we update that? (or else, there might be a case that someone misunderstood and try to validate it with `$this->applyToHooks([]) === false` which will be wrong)

jonrandy: Re-read the above

jonrandy: I often write code that takes advantage or truthiness and falsiness

guzzilar: @jonrandy am I missing something? 🤔 
I was checking the type of the code at line 👉  `return $res;` and got `int(0)` and `int(1)`

guzzilar:
@jonrandy  fyi, your text-editor messed up with that `.xml` file again lol (at the latest commit https://github.com/omise/omise-prestashop/pull/62/commits/28226172a152656a5cd752bd55704a8917472684)

jonrandy:
It's not my text editor. PrestaShop itself is doing it

PR #68, opened 2018-08-16
Fix version of Prestashop from 1.6.19 to 16.1.19

hassadee:
Prestashop v.16.1.19 is the correct version.


jonrandy:
There is a typo here, although not quite what you anticipated. We had tested with 1.6.1.9, not 1.6.1.19

PR #71, opened 2018-11-24
Fix bank image icons filepath

hassadee:
Fix bank image icon filepath from  "/modules/omise/img/" to "{$urls.base_url}/modules/omise/img/".


jonrandy:
Thanks very much for this. I won't actually merge this though as I've made another commit to fix this in both Prestashop 1.6 and 1.7.

The fixes will appear in the next release.

Thanks again :)

PR #73, opened 2019-05-31
Sandbox warning

jonrandy:
#### 1. Objective

Add a highly visible warning on checkout page so that merchants will easily know if they have left their site in 'Sandbox' mode for payments

#### 2. Description of change

Some CSS and JS are conditionally added to the checkout page if 'sandbox mode' is switched on.

image

#### 3. Quality assurance

Install the plugin in both PrestaShop 1.6 and 1.7 - check that the warning appears/disappears as expected on the checkout page.

#### 4. Impact of the change

Merchants will be far less likely to accidentally leave their sites in sandbox mode

#### 5. Priority of change

Normal

#### 6. Additional notes

None

### DIFF
--- a/omise/views/templates/hook/omise_warning_message.tpl
+++ b/omise/views/templates/hook/omise_warning_message.tpl
@@ -1,0 +1,8 @@
+<script type="text/javascript">
+	// <![CDATA[
+	var omiseText = {
+		warning: "{l s='OMISE SANDBOX MODE' mod='omise'}",
+		linkText: "{l s='more info' mod='omise'}"		
+	};
+	//]]>
+</script>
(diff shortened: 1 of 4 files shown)


omise/views/templates/hook/omise_warning_message.tpl
```diff
@@ -0,0 +1,8 @@
+<script type="text/javascript">
+	// <![CDATA[
+	var omiseText = {
+		warning: "{l s='OMISE SANDBOX MODE' mod='omise'}",
+		linkText: "{l s='more info' mod='omise'}"		
+	};
+	//]]>
+</script>
```
jacstn, line 8: you could consider adding break line...

jacstn (approved):

PR #74, opened 2019-09-03
Adds support for more currencies

jonrandy:
#### 1. Objective

Add support for the additional currencies that have been launched on the backend recently.

```
TH:

Australian Dollar (AUD)
Swiss Franc (CHF)
Chinese Yuan (CNY)
Danish Krone (DKK)
Euro (EUR)
British Pound (GBP)
Hongkong Dollar (HKD)
Japanese Yen (JPY)
Singapore Dollar (SGD)
US Dollar (USD)

SG:

Australian Dollar (AUD)
Canadian Dollar (CAD)
Swiss Franc (CHF)
China Yuan (CNY)
Euro (EUR)
British Pound (GBP)
Hongkong Dollar (HKD)
Japanese Yen (JPY)
Malaysian Ringgit (MYR)
Thai Baht (THB)
US Dollar (USD)
```

#### 2. Description of change

Modifications to the functions that:

a) Convert amounts to subunits
b) Check if a given currency is supported

#### 3. Quality assurance

Tested on my test sites for PrestaShop 1.6 and 1.7 - attempting payments in the new currencies

#### 4. Impact of the change

Newly added currencies can be used

#### 5. Priority of change

Normal

#### 6. Additional notes

👹

### DIFF
--- a/omise/libraries/omise-plugin/helpers/charge.php
+++ b/omise/libraries/omise-plugin/helpers/charge.php
@@ -130,7 +130,7 @@ if (! class_exists('OmisePluginHelperCharge')) {
         public static function isCurrentCurrencyApplicable($currency)
         {
             $curr = strtoupper($currency);
-            return array_key_exists($curr, self::$supportedCurrencies);
+            return in_array($curr, self::$supportedCurrencies);
         }
 
         /**
--- a/omise/libraries/omise-plugin/helpers/charge.php
+++ b/omise/libraries/omise-plugin/helpers/charge.php
@@ -63,7 +63,7 @@ if (! class_exists('OmisePluginHelperCharge')) {
         /**
          *
          * @param string $currency
-         * @param $amount
+         * @param numeric $amount
          * @return string
          */
         public static function amount($currency, $amount)
--- a/omise/libraries/omise-plugin/helpers/charge.php
+++ b/omise/libraries/omise-plugin/helpers/charge.php
@@ -2,25 +2,73 @@
 if (! class_exists('OmisePluginHelperCharge')) {
     class OmisePluginHelperCharge
     {
+
+        private static
+            $currencyDecimals = [
+                "BIF" => 0,
+                "CLP" => 0,
+                "DJF" => 0,
+                "GNF" => 0,
+                "ISK" => 0,
+                "JPY" => 0,
+                "KMF" => 0,
+                "KRW" => 0,
+                "PYG" => 0,
+                "RWF" => 0,
+                "UGX" => 0,
+                "UYI" => 0,
+                "VND" => 0,
+                "VUV" => 0,
+                "XAF" => 0,
+                "XOF" => 0,
+                "XPF" => 0,
+                "BHD" => 3,
+                "IQD" => 3,
+                "JOD" => 3,
+                "KWD" => 3,
+                "LYD" => 3,
+                "OMR" => 3,
+                "TND" => 3,
+            ],
+            $defaultCurrencyDecimals = 2,
+            $upportedCurrencies = [
+                'AUD',
+                'CAD',
+                'CHF',
+                'CNY',
+                'DKK',
+                'EUR',
+                'GBP',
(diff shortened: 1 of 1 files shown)


omise/libraries/omise-plugin/helpers/charge.php
```diff
@@ -2,25 +2,73 @@
...
+                "GNF" => 0,
+                "ISK" => 0,
+                "JPY" => 0,
+                "KMF" => 0,
+                "KRW" => 0,
+                "PYG" => 0,
+                "RWF" => 0,
+                "UGX" => 0,
+                "UYI" => 0,
+                "VND" => 0,
```
jacstn, line 20: why all those currencies are listed since they are not supported?

jonrandy: The list does not relate to supported currencies. It is used in a generic function to get the number of subunits for a given currency. I just added those so the function is very future proof

jacstn: why no `PLN` polish złoty, why no `Russian Rubel`? But `Uganda Shilling` is there?

jacstn:
@guzzilar, @jonrandy I am implementing that in `mag2`
different currencies are supported by different account country origin.
any ideas on how to solve that?

omise/libraries/omise-plugin/helpers/charge.php
jonrandy: Explained in the code. Doesn't need to be there if it has 2 decimal places

jacstn: ok, so what it does, it handles all currencies in the world, right?

jonrandy: Assuming the list I pulled the data from was correct - yes - at least until new currencies are added, but even then the likelihood is the new currency will probably have 2 decimal places like the majority of others

jacstn: ok... what are the chances that those currencies will be added by omise like UGX,
if you want to leave that in the code ok.
But that most likely will never, ever be used, ever...

jacstn (approved):

omise/libraries/omise-plugin/helpers/charge.php
```diff
@@ -2,25 +2,73 @@
...
+        public static function subUnitsFor($curr) {
+            $c = strtoupper($curr);
+            $decimals = isset(self::$currencyDecimals[$c]) ? self::$currencyDecimals[$c] : self::$defaultCurrencyDecimals;
+            return pow(10, $decimals);
+        }        
+
         /**
          *
          * @param string $currency
-         * @param integer $amount
```
jacstn, line 8: we could correct comment maybe `@param float $amount` ?

omise/libraries/omise-plugin/helpers/charge.php
jonrandy: Yeah, I was conflicted about whether to put them all in, but eventually decided it was better to have the function work properly for as many cases as possible without being wrong, or having to return something like `null` for a currency it didn't know. You're correct though, most are edge cases that will almost certainly never be used

omise/libraries/omise-plugin/helpers/charge.php
jonrandy: It accepts any var that can be 'used' as a number - I guess that would be mainly int/float/numeric string - maybe should just write `@param numeric $amount`

jonrandy:
> @guzzilar, @jonrandy I am implementing that in `mag2`
> different currencies are supported by different account country origin.
> any ideas on how to solve that?

Ultimately we need to use the capability API for it I think - each payment method lists the currencies it can use. One would assume these will be accurate

jacstn:
will not happen anytime soon - getting data in capabilities.

jonrandy:
It's already there. PR is awaiting merge

PR #78, opened 2020-01-23
Refactor

jonrandy:
#### 1. Objective

The old Prestashop plugin code had a lot of areas where code for numerous different payment methods was mixed up together. Adding new payment methods would entail sprinkling a bit of new code everywhere. It was generally quite messy.

#### 2. Description of change

A fairly radical overhaul of the code to attempt to group code for individual payment methods as much as possible. There's still work to do (admin area, and maybe relocating some template files) but this PR is already a monster.

Some minor bugs that had gone unnoticed are also fixed.

#### 3. Quality assurance

Re-check all payment methods still work, as well as changing/saving settings in admin. Needs to be tested on both versions of Prestashop - 1.6 and 1.7.

#### 4. Impact of the change

The codebase is now smaller and hopefully easier to deal with. It should also now be simpler to add  new payment methods

#### 5. Priority of change

Normal

#### 6. Additional notes

C, A#


guzzilar:
@jonrandy just wonder about the nature of PrestaShop file structure & naming convention. 
PrestaShop doesn't have some sort of naming convention on its filename?
i.e. in many platforms and PHP would say "filename should be reflected to its class".
`classs OmiseOffsitePaymentMethod` would become `class-omise-offsite-payment-method.php` in WordPress (for WP, there is a `class-` prefix).

jonrandy:
There don't seem to be any hard and fast rules... just some loose guidelines (which had already been ignored/overlooked by the time I inherited the plugin - way before these PRs). When using/creating PrestaShop 'core' stuff like Controllers and the like, I've followed the conventions I could see. Other than that I've just gone with what I thought best

guzzilar:
I see.
Reading from the original code. There are some classes that were, kinda ignored the naming pattern but I still see the intention of the author that try to stick with the "filename reflects to its class" (in `classes`, `events` folders). Except for those controller-front classes (however, still understandable that those classes are following the pattern of `filename_folder_structure`).

If there is no particular reason or benefit of breaking that style, then I think it's better to follow that original intention. It would make other readers catch up with the project structure faster. Also it will benefit in case we write some automation or even a test-suite later on.

Also with the above reason, I think the classes inside `omise/payment_methods/_all.php` should be separated into 2 files too. For the sake of automation-test.

guzzilar:
Also, I think we must run the test too as this project has already been implemented the uni-test.

jonrandy:
My intention is to as far as possible, separate the payment methods from the rest of the code. I have ideas to move more stuff into the payment methods folder.

The next PR removes the entire test suite. The codebase has changed so much that the tests were of no use. I have never in my career written unit tests for my code. I don't use them. Someone else is welcome to write some if they find it beneficial

guzzilar:
> My intention is to as far as possible, separate the payment methods from the rest of the code. I have ideas to move more stuff into the payment methods folder.

I think it is a good idea. No issue about this

omise/payment_methods/alipay.php
```diff
@@ -0,0 +1,28 @@
...
+
+class OmisePaymentMethod_Alipay extends OmiseOffsitePaymentMethod
+{
+
+    const
+        NAME = 'Alipay',
+        PAYMENT_OPTION_NAME = 'omise-alipay-payment',
+        DEFAULT_TITLE = 'Alipay',
+        TEMPLATE = 'alipay_payment',
+        CONTROLLER = 'alipaypayment'
```
mayurkathale: `const CONTROLLER` need to remove this if never used.

omise/payment_methods/internet_banking.php
```diff
@@ -0,0 +1,30 @@
...
+
+class OmisePaymentMethod_InternetBanking extends OmiseOffsitePaymentMethod
+{
+
+    const
+        NAME = 'InternetBanking',
+        PAYMENT_OPTION_NAME = 'omise-internet-banking-payment',
+        DEFAULT_TITLE = 'Internet Banking',
+        TEMPLATE = 'internet_banking_payment',
+        CONTROLLER = 'internetbankingpayment'
```
mayurkathale: Same

omise/payment_methods/alipay.php
jonrandy: Already removed in a later PR

omise/payment_methods/internet_banking.php
jonrandy: Already removed in a later PR

omise/views/templates/hook/1.6/internet_banking_payment.tpl
```diff
@@ -4,16 +4,16 @@
...
           <div class="col-sm-12">
-            <form id="omiseInternetBankingCheckoutForm" method="post" action="{$link->getModuleLink('omise', 'internetbankingpayment', [], true)|escape:'html'}">
+            <form id="omiseInternetBankingCheckoutForm" method="post" action="{$action|escape:'html'}">
               <ul class="omise-internet-banking">
                 <li class="item">
                   <input class="no-uniform" id="omiseInternetBankingScb" name="offsite" type="radio" value="internet_banking_scb" autocomplete="off">
                   <label for="omiseInternetBankingScb">
                     <div class="omise-logo-wrapper scb">
-                      <img src="{$urls.base_url}/modules/omise/img/scb.svg" class="scb">
+                      <img src="{$base_dir}/modules/omise/img/scb.svg" class="scb">
```
guzzilar, line 16: Do we wanna replace this one in the `1.7/internet_banking_payment.tpl`?

omise/setting.php
```diff
@@ -6,62 +6,82 @@
...
 
     /**
-     * @return string
+     * Add setting names to list to be saved/used
      */
-    public function getLivePublicKey()
+    public function addUsedSettings($settings)
     {
-        return $this->getConfig('live_public_key');
+        $this->all_settings = array_unique(array_merge($this->all_settings, $settings));
```
guzzilar, line 72: wonder in what case that we will use `array_unique` here?

omise/setting.php
```diff
@@ -6,62 +6,82 @@
...
-            'internet_banking_status',
-            'alipay_status'
+            'title'
+        )
+    ;
+
+    protected
+        $callToCFG = array(
+            // retrieve normal xxx_yyy_zzz config values by calling getXxxYyyZzz
+            array( 'match'=>'%^get[A-Z].*$%', 'find'=>array('%([A-Z])([a-z])%', '%get_%'), 'repl'=>array('_\1\2', '') ),
```
guzzilar, line 29: As this PR is implementing the `__call()` method, wonder if this could be replaced with a `__get()` magic method instead? Is there any reason that we want to execute a method instead of accessing a property from the getter method?

Just seems more simpler.

omise/payment_methods/alipay.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+if (defined('_PS_MODULE_DIR_')) {
+    require_once _PS_MODULE_DIR_ . 'omise/payment_methods/_offsite.php';
+}
+
+class OmisePaymentMethod_Alipay extends OmiseOffsitePaymentMethod
```
guzzilar, line 7: Understand that the `_` is on a purpose of automation, but I think we should strict with our project naming convention (it's not clearly set but I think the original approach is just simple camelCase without underscore).

omise/payment_methods/_all.php
```diff
@@ -0,0 +1,78 @@
+<?php
```
guzzilar, line 1: I don't know how to put this, but I think it is kinda standard in a modern PHP already that we should have one class per file. This gives a bit of strange look. (I don't know how to explain 'why' it feels that way)

omise/payment_methods/_all.php
```diff
@@ -0,0 +1,78 @@
...
+
```
guzzilar, line 2: Also, the same reason as above, it gives a bit of strange look that the classname and its filename is not reflect to each other. (cannot tell why it feels that way)

guzzilar (commented):
I've left some minor comments. 
After reviewed, with the understanding of a new structure. I have a concern about this approach, although it is very smart way and reduces a lot of duplication. I feel it's far more complicated than it needed to be in order to achieve the goal of removing redundant and automating, personally I feel it sacrifices the maintaining ability.

The original approach was too redundant and misplaced, agreed. But I think it could be a little less smart but more easy to maintain and decoupled.

However, just personal style and should not be any problem of approving the PR (besides those minor points).

(would it be weird to say "a little less smart code is better"? 😂)

omise/views/templates/hook/1.6/internet_banking_payment.tpl
jonrandy: Nope - I'd accidentally rolled this fix into both before, when the fix was only needed for 1.7. Hence the reversion here

omise/setting.php
jonrandy: Each payment method lists what settings it uses. We need to compile a list of all settings this settings class needs to be aware of. It's entirely possibly two different payment methods might need to use the same setting. If we didn't use `array_unique` here we'd end up with that name twice in the array

omise/setting.php
jonrandy: It could indeed be replaced in the future. It was better at the point in time when this was implemented however to stick with `__call` so that all legacy code that was calling the `getXXXXX` functions would still work

omise/payment_methods/alipay.php
jonrandy: Nothing to do with automation. It is my convention to let us know that the file contains merely a parent/base/utility classes, not a payment method that will actually be used. The underscores make the look different and keep them at the head of a file list

omise/payment_methods/_all.php
jonrandy: This is very minor - I'll split it in the current branch I'm working on if it bothers you

omise/payment_methods/alipay.php
jonrandy: I may have misunderstood - are you referring to file names or class names? If class name then I'd like to keep them as is. It was a conscious choice for readability - `OmisePaymentMethod_NameHere`

guzzilar: Yup I was mentioning about its class name. (nothing to add more, just reply to your question)

omise/payment_methods/_all.php
jonrandy: I've split it in my current branch

omise/payment_methods/_all.php
guzzilar: I see this point has been replied at another comment, so I just refer to that comment here instead
> It is my convention to let us know that the file contains merely a parent/base/utility classes, not a payment method that will actually be used. The underscores make the look different and keep them at the head of a file list

if that so, wouldn't be better if (just idea) instead we make it to be abstract class and name it accordingly?
I see the convention that has been implemented here is reflect to its directory and its filename. 
`omise/payment_method/alipay` becomes `OmisePaymentMethod_Alipay`.
So just feel it's strange when we introduce that `_all.php` and `_offsite.php`. 

(it's minor so I'd like to leave it to you, just share my thought)

guzzilar (approved):
I think this approach is a bit too complicated than it needed to be (as mentioned earlier about maintainability). However, just personal taste & style.

Thanks for the PR! : D

guzzilar:
Btw just fyi, because PR #79 is submitting the changes to this PR. So if this get merged and deleted then I believe that all those related 5 PRs will be automatically closed.
(normally if there is a stack PR like this, I'll submit it all to the `master` branch and rebase one by one once its parent PR get merged)

jonrandy:
My intention is to merge the stack of PRs from the other end, so it's not an issue

PR #80, opened 2020-01-29
Citi

jonrandy:
#### 1. Objective

Adds Citipoints payment to Prestashop

#### 2. Description of change

Added the few files necessary to add the new payment method (proves the refactoring was worthwhile! 😄 ). Updated OmisePHP and updated API version (was necessary to allow Citi to work) .

#### 3. Quality assurance

Checked all admin and payment function relating to Citi worked ok. Also made sure all previous payment methods still working (due to the OmisePHP and API version changes)

#### 4. Impact of the change

Payment with Citi points now available in Prestashop!

#### 5. Priority of change

Normal


omise/classes/omise_charge_class.php
```diff
@@ -50,7 +50,7 @@ public function create($card_token)
     }
 
     /**
-     * @param string $offsite The parameter used to specify a bank to create Omise internet banking charge.
+     * @param string/array $offsite The parameter used to specify simple offsite type, or array of type and additional params
```
mayurkathale, line 53: Documentation comment need to change here.
`* @param string | array $offsiteType` or ` *@param mixed $offsiteType`

jonrandy: ? I've said it can be a string or array

mayurkathale (dismissed):
Requesting minor change. Please see notes.

omise/classes/omise_charge_class.php
mayurkathale: Right but string|array should be written in docs (not with /)

jonrandy: I had no idea. We're not strict on this stuff anyway - you'll probably find loads of stuff like that around. Changed in my latest branch 👍

omise/controllers/front/base.php
```diff
@@ -19,7 +19,7 @@
  * @see OmiseApiResource::genOptions() The function in library, Omise PHP, that uses this constant.
  */
 if (! defined('OMISE_API_VERSION')) {
-    define('OMISE_API_VERSION', '2015-11-17');
+    define('OMISE_API_VERSION', '2017-11-02');
```
guzzilar, line 22: just wonder if there is any reason we don't go with the latest API version (2019)?

guzzilar (commented):
@jonrandy could you add screenshots of this change as well? as it provides new UIs to the plugin : )

omise/controllers/front/base.php
jonrandy: No real reason - I copied it from another plugin where it was used and everything was known to be working

jonrandy:
> @jonrandy could you add screenshots of this change as well? as it provides new UIs to the plugin : )

Not worth it - just adds an on/off switch identical to that from other payment methods. UI changes are self explanatory

guzzilar:
> Not worth it - just adds an on/off switch identical to that from other payment methods. UI changes are self explanatory 

Worth it for what?. The changes have modified the admin page and introduce a new payment at the check out page. This is just a normal code review process of providing details of [2. description of change]

jonrandy:
I think images should only ever be necessary if changes are hard to spot or not obvious. If the PR states it’s adding a payment method and says it is changing/adding an admin template it should be obvious where the changes will appear - at the checkout and on the admin page. As for showing what they look like at the front end - this is dependent upon the theme being used on the store - which could be different between sites - so the changes in the files give a better indication of the change - i.e. the HTML will show any new fields etc.

guzzilar:
> I think images should only ever be necessary if changes are hard to spot or not obvious.

I think it is helping on a better & clear communication on what changes because reviewers or anyone that read this pull request may have a different level of understanding on the PrestaShop UI and what has been added.

But if you think it is not necessary then ok.

guzzilar (approved):

PR #82, opened 2020-02-04
Payment method JS cleanup

jonrandy:
#### 1. Objective

Clean up the JS in the payment method templates

#### 2. Description of change

Small refactoring of the JS on the card and internet banking methods - reduces repeated code and unduly verbose code

#### 3. Quality assurance

Re-tested associated payment methods on both PS 1.6 and 1.7

#### 4. Impact of the change

JS is easier to maintain, a little faster to load as smaller

#### 5. Priority of change

Normal


mayurkathale (approved):
LGTM

omise/views/templates/hook/1.6/card_payment.tpl
```diff
@@ -77,63 +77,49 @@
...
+    'name',
+    'number',
+    'expiration_month',
+    'expiration_year',
+    'security_code'
+  ];
+
   window.omiseCheckout = function omiseCheckout() {
     if (typeof Omise === 'undefined') {
       alert('{l s='Unable to process the payment, loading the external card processing library is failed. Please contact the merchant.' mod='omise'}');
```
guzzilar, line 90: Not sure if this is also in a scope of JS cleanup but wonder if we should change the outer single quote to be double quote instead? 
```js
alert("{l s=' blah ' mod='omise'}");
```

guzzilar: or I guess it is a PrestaShop magic...

omise/views/templates/hook/1.6/card_payment.tpl
```diff
@@ -77,63 +77,49 @@
...
-  window.omiseLockCheckoutForm = function omiseLockCheckoutForm(form) {
-    form.name.disabled = true;
-    form.number.disabled = true;
-    form.expiration_month.disabled = true;
-    form.expiration_year.disabled = true;
-    form.security_code.disabled = true;
-    form.checkout_button.disabled = true;
-    form.checkout_text.innerHTML = '{l s='Processing' mod='omise'}';
-  };
+  window.omiseSetCardFormLockedState = function omiseSetCardFormLockedState(form, state) {
```
guzzilar, line 120: I had a very long long time relationship with JS and it's not quite up-to-date.
Wonder do we need to declare function name again? 
Can it be `window.omiseSetCardFormLockedState = function(form, state) { ... }`?

jonrandy: It was like that already. I believe it assists in debug trace. If you don't give a name to the function, they just appear anonymously in the stack. Your example creates a nameless function and assigns it to a variable, which is different to declaring a named function

omise/views/templates/hook/1.7/card_payment.tpl
```diff
@@ -65,74 +65,53 @@
...
+        var card = omiseCardFields.reduce(function(obj, field) { obj[field] = form[field].value; return obj;}, {});
+        Omise.setPublicKey('{$omise_public_key}');
+        Omise.createToken('card', card, omiseCreateTokenCallback);
+      },
+
+      isOmiseCardPaymentOptionSelected = function isOmiseCardPaymentOptionSelected() {
+        return document.querySelector('[data-module-name="omise-card-payment"]').checked;
+      },
+
+      setPayFormLockState = function setPayFormLockState(form, state) {
```
guzzilar, line 97: Wonder if the name would be better with `setCardFormLockedState` the same as in template 1.6

guzzilar (approved):
Some minor comments.
However, I learned a lot from this refactoring.
Interesting one 👍

PR #83, opened 2020-02-18
Offsite refactor + relocating 'returnURI' processing

jonrandy:
#### 1. Objective

Further refactor the offsite methods to avoid duplicated code and ease creation of new payment methods. Also move all processing done by 'return' controller into paymentMethod class

#### 2. Description of change

- Made offsite methods share code and templates where appropriate
- Moves Card payment processing code inside its payment method class from the controller
- Moves 'return' processing from Controller into payment method class
- Cleanup - move some objects from controllers into payment method classes where appropriate

#### 3. Quality assurance

Re-tested all payment methods on both versions of PrestaShop

#### 4. Impact of the change

Cleaner code structure, and should now be easier again to add Payment methods going forward

#### 5. Priority of change

Normal

#### 6. Additional notes

Killer whales are actually dolphins 🐬


omise/classes/omise_charge_class.php
```diff
@@ -62,7 +62,7 @@ public function createOffsite($offsiteType)
             'description' => $this->getChargeDescription(),
             'metadata' => $this->getMetadata(),
             'source' => is_array($offsiteType) ? $offsiteType : array('type' => $offsiteType),
-            'return_uri' => $this->getReturnUri(),
+            'return_uri' => $returnUri
```
mayurkathale, line 65: $returnUri is undefined.

jonrandy: Yes it is, it's a parameter to the function. There are cases where it is not needed

mayurkathale: In case of, offsite payments there should be some value, right?
right now it looks like it will be always NULL for all offsite payments.

jonrandy: Already fixed in another later commit d173bfb6a94c3a879f8d8e7be8e1a7fba5e98790

mayurkathale (approved):

guzzilar:
Seeing the changes up until this PR, I'm wondering about the idea behind of refactoring the `payment_methods/_all.php`. It seems we are relocating the most, if not all, **payment-related** functions out from places and move it to `OmisePaymentMethod` class in `payment_methods/_all.php`. 

I just see that `OmisePaymentMethod` becoming to take care of many stuff and getting bigger and bigger. Can you help explain a bit of the idea on why are we heading this way? Why having a **"one-for-all"** (OmisePaymentMethod) class is better than the current (previous) approach?

guzzilar:
(It just seems we are twisting lots of code so the `payment_methods/_all.php` concept can work. But now this class seems to hold lots of responsibilities (seeing from its methods))

jonrandy:
It's a payment method. It needs to be able to do certain things and handle certain things. These all relate to the payment method. Feels way better to have them all in one place instead of spread through different classes. The OmisePaymentMethod class in the `_all.php` file contains those things that are common to all payment methods. Other methods inherit from this class.

Just feels much cleaner this way than to have various functions of a payment method scattered through controllers etc.

guzzilar:
I agree that a payment method class should be concerning and knowing how to handle things, but up to a certain degree. I just think this is a bit code spaghetti.

For example, `OmisePaymentMethods::getValidCharge()`. 
We have `OmiseChargeClass` already, why `OmisePaymentMethods` class needs to care how to retrieve a valid charge, why not let `OmiseChargeClass` class itself to handle that task.

Or, `OmiseChargeClass::create()` is already coupled with `Context` class ($this->context), now we're relocating `getReturnUri` out to `OmisePaymentMethods` meaning that these 2 classes are involving with `Context`. I feel this makes the code more complicated.

Or, we are relocating the part that handling the return payment to `OmisePaymentMethods:: handleReturn()`. We have controller structure to handle the context of how to handle result (redirection, handle error message, etc.) and now we are leaking controller context into `OmisePaymentMethods` class.

Understand that `OmisePaymentMethods` is like a model, or a module that is handling communication between these 3 interfaces: API <> Omise plugin <> PrestaShop.
But leaking the controller context into this class and now this class has to know how to handle everything. It just seems started to become a god class. A bit too spaghetti and too coupled.  

Also, from `OmisePaymentMethods::getOrderConfirmationUri()`
Now the `OmisePaymentMethods` has to concern about the url structure of PrestaShop.
It doesn't seem like what payment method class should be concerning.

guzzilar (approved):
With the following reasons https://github.com/omise/omise-prestashop/pull/83#issuecomment-603432681, I still respectfully disagree on this approach (only PR #83). Mainly I feel this approach make code more complicated than it needed to be. Also, giving too much responsibility on one class (the OmisePaymentMethod class) and I just feel some of the code shouldn't belong there.

However, as discussed internally and as I don't have another alternative idea to propose, it's time to move this refactoring on.

Issue #89, opened 2020-04-07
Update from 1.7.6 to 1.7.8 Module payment not working anymore


noobookbig:
On Prestashop 1.7.6.4 with module 1.7.6 are working when we have updated to 1.7.8 module is not working correctly. when user selects any payment and click on pay the page will be redirect to itself which happened on sandbox and live mode.

jonrandy:
I have not been able to reproduce this. Does this happen on all browsers? Are their any errors?

noobookbig:
I have got only info from the admin log

Informative only (1) | Frontcontroller::init - Cart cannot be loaded or an order has already been placed using this cart | Cart | 67 | 0 | 2020-04-07 11:29

Yes, it has happened on all browser. 
It does not show any error on the page when debug mode is enabled so it reloads the page itself. Try to clear cache does not solve it also.

jonrandy:
Did you clear both the browser cache and all of PrestaShop's caches?

noobookbig:
Yes I have clear all cache.

jonrandy:
Do you have any other PrestaShop plugins active?

noobookbig:
I have only the theme plugins.

![screencapture-zugarmellow-admin500683iz7-index-php-improve-modules-manage-2020-04-07-12_12_19](https://user-images.githubusercontent.com/12130745/78632388-14be5900-78c9-11ea-9f3e-8445de7b2908.png)

jonrandy:
Is it possible I can look at the site? If so, you can send the details to jonathan@omise.co

noobookbig:
Ok Information sent to your email

noobookbig:
Dear Jon, I have found the issue.


![ssl](https://user-images.githubusercontent.com/12130745/78764642-0a778a00-79b1-11ea-90ab-6fc480a1c803.JPG)
The issue caused by there are some insecure endpoint which is using HTTP request however on Prestashop setting I have to enable Force SSL on all page, I have workaround to get this working by not use strict mode on Cloudflare it will replace HTTP request by https and using Cloudflare certificate.

jonrandy:
OK. I've made a small change to remove the need for the workaround. Thanks again for helping with this

PR #90, opened 2020-04-09
Fix small HTTPS issue and prepare 1.7.9 release

jonrandy:
#### 1. Objective

**Related information:**
Related issue(s): #89 

This fixes a small issue (possibly introduced during refactoring) that was causing issues in environments where HTTPS was being enforced everywhere (no payment methods would work at checkout, customers were just redirected back to the checkout page).

The PR also makes the small preparations necessary for v1.7.9 release with this fix.

#### 2. Description of change

The function responsible for generating links was changed to default to secure.

#### 3. Quality assurance

I worked with the merchant who reported the issue to get him to test it in the environment where the issue was occurring. He confirmed it now worked fine.

#### 4. Impact of the change

Plugin will now work successfully in strict HTTPS only environments

#### 5. Priority of change

Normal

### DIFF
--- a/omise/payment_methods/_payment_method.php
+++ b/omise/payment_methods/_payment_method.php
@@ -72,7 +72,7 @@ class OmisePaymentMethod
         return count(static::$restrictedToCurrencies) ? in_array($curr, static::$restrictedToCurrencies) : true;
     }
 
-    public static function getLink($method, $params = [], $controller = 'paymentmethod', $ssl = false) 
+    public static function getLink($method, $params = [], $controller = 'paymentmethod', $ssl = true) 
     {
         return self::$context->link->getModuleLink(Omise::MODULE_NAME, $controller, array_merge($params, array('type' => $method)), $ssl);
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## [[1.7.9] 2020-04-09](https://github.com/omise/omise-prestashop/releases/tag/v1.7.9)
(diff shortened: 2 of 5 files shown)


omise/payment_methods/_payment_method.php
```diff
@@ -72,7 +72,7 @@ public static function availableForCurrency($code)
         return count(static::$restrictedToCurrencies) ? in_array($curr, static::$restrictedToCurrencies) : true;
     }
 
-    public static function getLink($method, $params = [], $controller = 'paymentmethod', $ssl = false) 
+    public static function getLink($method, $params = [], $controller = 'paymentmethod', $ssl = true)
```
guzzilar, line 75: Should the default value be `null` instead? Or is there any particular reason why do we want the default to be `true` (I mean, yeah `$ssl = true` is solving the problem, but just wonder why not leave it to PrestaShop setting)? (just checked the `getModuleLink` method from PrestaShop's `LinkCore` class, it leaves its default value as null)

jonrandy: I did consider that, so it would presumably follow the prestashop settings, but I deferred to what the plugin was doing before the refactor, which was making all the payment method links HTTPS. I thought that was safer

guzzilar: okay

guzzilar (approved):
I think we should reference to a related issue ticket in the PR's description as well.
In another repo I use template:

```
### 1. Objective

**Related information:**
Related issue(s): GitHub Ticket / Internal ticket
```

Just to make it easier to track back to its original request/issue

jonrandy:
Done

PR #98, opened 2023-12-19
Update README.md

muthuswamyopn:
## Description

Updated Omise to Opn Payments.


muthuswamyopn:
@danfowler  Do we need the Omise logo? It doesn't match the text then - text is modified to Opn Payments instead of Omise.

danfowler (approved):

aashishgurung (approved):

